PR #2, opened 2024-12-16
Update package name across the project

e5l:
Change package, update related references, docs, and publication.

The only unchanged folder is samples (needs to be fixed after the first release)


settings.gradle.kts
```diff
@@ -1,5 +1,5 @@
 plugins {
     id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
 }
-rootProject.name = "kotlinx-mcp-sdk"
+rootProject.name = "kotlinx-sdk"
```
Mr3zee: `kotlin-sdk`

PR #15, opened 2024-12-17
Setup sign and publishing configuration

e5l:
Setup GitHub action and gradle configuration to sign publish artifacts

### DIFF
--- a/.github/workflows/gradle-publish.yml
+++ b/.github/workflows/gradle-publish.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Build with Gradle
         run: ./gradlew build
 
-      - name: Publish to GitHub Packages
+      - name: Publish to OSSRH repository
         run: ./gradlew publish
         env:
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}


.github/workflows/gradle-publish.yml
```diff
@@ -0,0 +1,47 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+# This workflow will build a package using Gradle and then publish it to GitHub packages when a release is created
+# For more information see: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#Publishing-using-gradle
+
+name: Gradle Package
```
Mr3zee: `Release`?

.github/workflows/gradle-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          server-id: ossrh # Value of the distributionManagement/repository/id field of the pom.xml
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_PASSWORD
```
Mr3zee: `OSSRH_USERNAME` and `OSSRH_TOKEN`

e5l: please check: https://docs.github.com/en/actions/use-cases-and-examples/publishing-packages/publishing-java-packages-with-maven#publishing-packages-to-the-maven-central-repository

.github/workflows/gradle-publish.yml
e5l: thank you, fixed

e5l:
Hey, @Mr3zee! Thank you for the review, pushed update

.github/workflows/gradle-publish.yml
Mr3zee: Yes, but you have `OSSRH_USERNAME` and `OSSRH_TOKEN` defined in gradle configs and in secrets mapping in this yaml below

Mr3zee:
```yaml
env:
  OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
  OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
```

Mr3zee:
```kotlin
val ossrhUsername = System.getenv("OSSRH_USERNAME")
    ?: project.findProperty("kotlin.mcp.sdk.packages.username") as String?
val ossrhPassword = System.getenv("OSSRH_TOKEN")
    ?: project.findProperty("kotlin.mcp.sdk.packages.password") as String?
```

Mr3zee: They should be unified to use one name, I just think `OSSRH_` ones are better

e5l: Thanks, aligned!

.github/workflows/gradle-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          server-id: ossrh # Value of the distributionManagement/repository/id field of the pom.xml
+          server-username: OSSRH_USERNAME
+          server-password: OSSRH_TOKEN
```
dsp-ant, line 32: Will this read the username and token already or just insert it into the pom.xml?

.github/workflows/gradle-publish.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: ./gradlew build
+
+      - name: Publish to GitHub Packages
+        run: ./gradlew publish
+        env:
+          OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
+          SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
+          SIGNING_KEY_PRIVATE: ${{ secrets.SIGNING_KEY_PRIVATE }}
+          SIGNING_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
```
dsp-ant: If only this step needs the secrets, than we should do this:

```suggestion
      - name: Publish to GitHub Packages
        run: ./gradlew publish
        environment: release
        env:
          OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
          OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
          SIGNING_KEY_ID: ${{ secrets.SIGNING_KEY_ID }}
          SIGNING_KEY_PRIVATE: ${{ secrets.SIGNING_KEY_PRIVATE }}
          SIGNING_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
```

See https://github.com/modelcontextprotocol/python-sdk/blob/main/.github/workflows/publish-pypi.yml where we only call it in the release step. This way only the actually publishing requires a review.

.github/workflows/gradle-publish.yml
e5l: It will read it from the env section below

.github/workflows/gradle-publish.yml
e5l: sure, let me adjust it

e5l: The schema suggests that the property is forbidden inside the step:
image

dsp-ant: Ahh because this is a step. We have different "jobs" that depend on each other in the pypi.yml.

Issue #11, opened 2024-12-17
Convert project to a KMP library


Mr3zee:
No major roadblocks here

e5l:
Done

PR #22, opened 2024-12-18
Migrate project to Multiplatform (KMP)

devcrocod:
related to #11 

## How Has This Been Tested?
Tested with JVM tests and verified locally that it builds for native targets (ios)

## Breaking Changes

Updated the API for stdio: stdio now expects `Source` and `Sink` as input instead of `InputStream` and `OutputStream` respectively

## Changes
- updated `Dokka` version to "2.0.0"
- added `atomicfu`. Replaced Atomic from java to Atomic from atomicfu
- replaced java.io with kotlinx-io throughout the project
- modified the build and publication for the multiplatform project


e5l (approved):
please rebase before merging

e5l:
Let me rebase!

Salamndr4n20 (commented):
Lit, good stuff

PR #31, opened 2024-12-22
Fix the difference in error handling between Kotlin and Typescript implementations in onClose

ignatov:
Makes the error handling behavior consistent between Kotlin and TypeScript implementations when handling connection closure events.


e5l (approved):
lgtm

PR #23, opened 2024-12-18
Modify SSE config for server

Mr3zee:
Modify SSE config for server:
- rename to Kotlin style guide
- remove NPE
- manage paths clearly

UPD: Additionally fixes compilation in master and hanged tests

PR #37, opened 2025-01-20
Add merge queue rules

Mr3zee:
Added Merge Queue Rules, so that all merged PRs are validated

### DIFF
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -1,0 +1,15 @@
+name: Before Merge
+
+on:
+  merge_group:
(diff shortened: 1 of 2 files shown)


e5l:
Could you tell me why validate PR was skipped?

e5l (commented):
please make sure workflow runs before merging

Mr3zee:
@e5l validate was skipped to omit running it on every commit pushed to a pr. However, it will be executed when put in a merge queue before merging.

It can not be tested here, because compilation is broken in main, I'll execute it in #23

PR #44, opened 2025-02-18
Enable auto-merge instead of merge queue

Mr3zee:
GH doesn't allow to run `workflow_dispatch` events on forks, and another type of event is needed. This added trigger for the auto-merge event, and should solve the problem


Mr3zee:
Well, at least Action triggered here correctly

e5l (approved):

Issue #42, opened 2025-02-11
mcp kotlin server does not process JSONRPCRequest with null param, but python sdk may send with null param


shendaxia-sm:
**Describe the bug**

this request works with kotlin mcp server:

{“id”:“1”,“jsonrpc”:“2.0”,“method”:“tools/list”,“params”:{}}

but this request not works:

{“id”:“1”,“jsonrpc”:“2.0”,“method”:“tools/list”}

python sdk client send with null params, but kotlin sdk deserialize json with default null param, and will not invoker handler when param lis null

so kotlin sdk should set params default to EmptyJsonObject

**To Reproduce**
use python SSE Transport client send list tool command to a kotlin mcp server, will reproduce this issue.

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

e5l:
Hey @shendaxia-sm, thanks for the report.

tomakehurst:
This looks like the same issue as https://github.com/modelcontextprotocol/kotlin-sdk/issues/56, which has a draft fix in PR #58

devcrocod:
This was fixed in #157

PR #46, opened 2025-02-21
Process JSONRPCRequest with default param (#42)

JsFlo:
Fixes issue #42 by setting the default to `JSONRPTCRequest.params` from null to `EmptyJsonObject`. This makes it so that the absence of params in a request is treated as an empty array which is what the Python sdk also does.

## Motivation and Context
This makes it so that these two request bodies are treated similarly where previously not including params would result in an empty response.

`{"jsonrpc": "2.0", "method": "tools/list","params": {}, "id": 1}`
and 
`{"jsonrpc": "2.0", "method": "tools/list", "id": 1}`

## How Has This Been Tested?
I tested this with the sample Kotlin server project (with a bit of modification to update sample to use more up to date code) and a basic python SSE client.

I would use the python client to get the sessionId and then I would run

```
curl -v -X POST "http://localhost:3001/message?sessionId=9799477a-945f-4498-ac8e-fc9156d5898f" \
     -H "Accept: text/event-stream" \
     -H "Content-Type: application/json" \
     -d '{"jsonrpc": "2.0", "method": "tools/list","params": {}, "id": 1}'
```
with `"params": {},` I would get the listed tools and without it I would get an empty response(`EmptyRequestResult`).

I also added a unit test that would fail before my changes (would be EmptyRequestResult) and pass after my changes
 

## Breaking Changes
No

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ?] Breaking change (fix or feature that would cause existing functionality to change)
    * this will technically change behavior from certain requests getting empty results to those requests getting correct results
- [ ] Documentation update

## Checklist
- [x ] I have added or updated documentation as needed

### DIFF
--- a/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
+++ b/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
@@ -440,7 +440,7 @@ public abstract class Protocol(
         val serializer = McpJson.serializersModule.serializer(requestType)
 
         requestHandlers[method.value] = { request, extraHandler ->
-            val result = request.params?.let { McpJson.decodeFromJsonElement(serializer, it) }
+            val result = McpJson.decodeFromJsonElement(serializer, request.params)
             val response = if (result != null) {
                 @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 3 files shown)


e5l (approved):
Nice catch, looks good

Issue #72, opened 2025-04-10
client: transport start failed will  make listTools stuck


phodal:
**Describe the bug**

Hi, according to README, I create follow code. When connect to a high version for MCP Java SDK (0.8.0), will stuck in connect, not run to listTools

```kotlin
val tools = try {
    client.connect(transport)
    val listTools = client.listTools()
    listTools?.tools?.forEach { tool ->
        toolClientMap[tool] = client
    }
    listTools?.tools ?: emptyList()
} catch (e: Exception) {
    logger<CustomMcpServerManager>().warn("Failed to list tools from ${entry.key}: $e")
    emptyList<Tool>()
}
```

**To Reproduce**
1.  create a Java server: https://github.com/unit-mesh/autodev-mcp-test/tree/master/java-mcp
2. use Kotlin to connect Java server

**Expected behavior**

Should compatibility with Other language SDK, it not should throw issue in start or listTools

**Logs**

My case should be the different Schema between Java and Kotlin version. So when the server use latest version, will not correct identify `params._meta` and  `params.method` from Kotlin version.

![Image](https://github.com/user-attachments/assets/6a692a9c-58f9-4718-8e57-29c275083133)

Here is the log

```bash
{"jsonrpc":"2.0","id":4,"result":{"protocolVersion":"2024-11-05","capabilities":{"logging":{},"tools":{"listChanged":true}},"serverInfo":{"name":"mcp-server","version":"1.0.0"}}}
{"id":4,"method":"list/tool","params":{"protocolVersion":"2024-11-05","capabilities":{"experimental":{},"sampling":{}},"clientInfo":{"name":"weather","version":"1.0.0"},"_meta":{},"method":"initialize"},"jsonrpc":"2.0"}
{"jsonrpc":"2.0","id":4,"error":{"code":-32601,"message":"Method not found: list/tool"}}
{"id":4,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{"experimental":{},"sampling":{}},"clientInfo":{"name":"weather","version":"1.0.0"},"_meta":{},"method":"initialize"},"jsonrpc":"2.0"}
[...]

PR #74, opened 2025-04-11
fix: client: transport start failed will make listTools stuck #72

iptton:
transport may be closed before send message, it should throw exception by `Transport#sendMessage` instead failed silently with safety call

## Motivation and Context

transport may be closed before send message, it should throw exception by `Transport#sendMessage` instead failed silently with safety call otherwise , `result.await()` will never return

## How Has This Been Tested?

it should be test by `ClientTest.kt`, but that may not necessary. it will happen when transport closed at the point `withTimeout` called, it's hard to simulate that case but you can remote the `val transport = this@Protocol.transport ?: throw ....` to simulate it...

```kotlin
@Test
    fun `should throw exception when transport closed before call list`() = runTest(timeout = 10.seconds) {
        val serverOptions = ServerOptions(
            capabilities = ServerCapabilities(
                resources = ServerCapabilities.Resources(null, null),
                tools = ServerCapabilities.Tools(null)
            )
        )
        val server = Server(
            Implementation(name = "test server", version = "1.0"),
            serverOptions
        )

        server.setRequestHandler<InitializeRequest>(Method.Defined.Initialize) { request, _ ->
            InitializeResult(
                protocolVersion = LATEST_PROTOCOL_VERSION,
                capabilities = ServerCapabilities(
                    resources = ServerCapabilities.Resources(null, null),
                    tools = ServerCapabilities.Tools(null)
                ),
                serverInfo = Implementation(name = "test", version = "1.0")
            )
        }

        server.setRequestHandler<ListResourcesRequest>(Method.Defined.ResourcesList) { request, _ ->
            ListResourcesResult(resources = emptyList(), nextCursor = null)
        }

        server.setRequestHandler<ListToolsRequest>(Method.Defined.ToolsList) { request, _ ->
            ListToolsResult(tools = emptyList(), nextCursor = null)
        }

        val (clientTransport, serverTransport) = InMemoryTransport.createLinkedPair()

        val client = Client(
            clientInfo = Implementation(name = "test client", version = "1.0"),
            options = ClientOptions(
                capabilities = ClientCapabilities(sampling = EmptyJsonObject),
            )
        )

        listOf(
            launch {
                client.connect(clientTransport)
            },
            launch {
                server.connect(serverTransport)
            }
        ).joinAll()

        // Server supports resources and tools, but not prompts
        val caps = client.serverCapabilities
        assertEquals(ServerCapabilities.Resources(null, null), caps?.resources)
        assertEquals(ServerCapabilities.Tools(null), caps?.tools)
        assertTrue(caps?.prompts == null) // or check that prompts are absent

        // simulate transport closed before get resource or tool list
        // this may happen when transport config incorrectly
        clientTransport.close()

        // These should not throw
        client.listResources()
        client.listTools()

        // This should fail because prompts are not supported
        val ex = assertFailsWith<IllegalStateException> {
            client.listPrompts()
        }
        assertTrue(ex.message?.contains("Server does not support prompts") == true)
    }
```

## Breaking Changes
NO

## Types of changes

- [ ] Documentation update

## Checklist


e5l (changes requested):
could you add the test in the codebase?

Issue #76, opened 2025-04-15
Support for annoation


alpeshvas:
**Is your feature request related to a problem? Please describe.**

Currently, to describe tool definition and description of tools, lots of verbose is needed, can be done with kotlin idiomic way using annoatation Kotlin annoatations.

**Describe the solution you'd like**
Kotlin annotation: https://kotlinlang.org/docs/annotations.html

**Describe alternatives you've considered**

**Additional context**

Instead of this: 

```
server.addTool(
    name = "get_forecast",
    description = """
        Get weather forecast for a specific latitude/longitude
    """.trimIndent(),
    inputSchema = Tool.Input(
        properties = buildJsonObject {
            putJsonObject("latitude") {
                put("type", "number")
            }
            putJsonObject("longitude") {
                put("type", "number")
            }
        },
        required = listOf("latitude", "longitude")
    )
)
```

```
@McpTool(
    name = "get_forecast",
    description = "Get weather forecast for a specific latitude/longitude",
)
fun getForecastTool(latitude: String, longitude: String) {
```

adamglin0:
Here's Koog's tool declaration based on the comments, which might be helpful. I haven't looked at it closely yet. （Just support jvm target.）
https://docs.koog.ai/annotation-based-tools/

Issue #70, opened 2025-04-09
When to add KMP support?


Stream29:
This MCP SDK implementation is built on ktor. I cannot see any reason that this project uses `kotlin("multiplatform")` plugin but only add the `jvm` target.

We developers are looking for a implementation on KMP that allows us use it on other platform.

alexstyl:
Afaik ktor server recently went multiplatform and can run on the web targets now. That might probably the reason why

devcrocod:
@Stream29 thank you for the issue
To better understand your use case: are you looking for a multiplatform client, server, or both? Knowing this will help us prioritize and plan the implementation accordingly.

alexstyl:
I am personally looking into adding my JVM + WASM app a MCP server

Stream29:
@devcrocod I'm working on a kmp version of chatbox, trying to add support of MCP with stdio transport, just like Claude Desktop app. 

After reading the source code of this implementation of mcp, I found most of its code is platform independent and can be migrated to other targets easily.

So I add this issue.

Issue #32, opened 2024-12-28
support generating json schema for tools


christophsturm:
the mcp docs for tools contain this:
Best practices

When implementing tools:
...
Use detailed JSON Schema definitions for parameters
..

and the typescript sdk examples just dynamically create json schema from the zod types that the typescript examples use. it would be great if the kotlin sdk has something like this too, or an example how to do it. (the current example is very bare bones)

morisil:
I'm the author of open source library providing this functionality:

https://github.com/xemantic/xemantic-ai-tool-schema

christophsturm:
@morisil how can i use that with the kotlin mcp sdk? the json schema field is of type Input:
```
    @Serializable
    public data class Input(
        val properties: JsonObject = EmptyJsonObject,
        val required: List<String>? = null,
    ) {
        val type: String = "object"
    }
```
or are you suggesting to fork the sdk? also your claudine library seems to have some overlap with mcp in general, are you planning to make it mcp compatible?

morisil:
As a quick workaround - add a function translating `JsonSchema` instance, as produced by my library, to the `Input` instance of this SDK Tool. Or even instantiate the whole `Tool` with description from annotated class, as I am doing in my `antrhopic-sdk-kotlin`:

https://github.com/xemantic/anthropic-sdk-kotlin#using-tools

A long term solution I would like to propose for this project - move my `xemantic-ai-tool-schema` under the umbrella of the mcp project, and use the `JsonSchema` class in the API. The argument for this is that such a `JsonSchema` can be used in several other SDKs, for other LLMs, avoiding boilerplate and translation layers.

morisil:
Regarding Claudine, I want to keep Claudine1 as an educational project showing the simplest Unix-omnipotent agent on top of the anthropic-sdk-kotlin, with multiplatform compilation targets. Claudine2 I am developing at the moment (might have a different name), will utilize MCP to a great extent.

morisil:
@christophsturm As a quick workaround you can use something like:

```kotlin
import com.xemantic.ai.tool.schema.ObjectSchema
import com.xemantic.ai.tool.schema.generator.jsonSchemaOf
import io.modelcontextprotocol.kotlin.sdk.Tool.Input
import kotlinx.serialization.json.Json
import kotlinx.serialization.json.encodeToJsonElement
import kotlinx.serialization.json.jsonArray
import kotlinx.serialization.json.jsonObject
[...]

PR #80, opened 2025-04-15
Support for MCP definations using kotlin annotation

alpeshvas:
## Motivation and Context

Add support for #76 

## How Has This Been Tested?

Unit tested and built jar was locally tested with MCP server built with annotations.

## Breaking Changes

No

## Types of changes

- [ ] Documentation update

## Checklist


alpeshvas:
@devcrocod @e5l it's ready for review.

devcrocod:
Sorry, I don’t fully understand the motivation.
There is already [xemantic-ai-tool-schema](https://github.com/xemantic/xemantic-ai-tool-schema), which was discussed in #32
There is also [victools/jsonschema-generator](https://github.com/victools/jsonschema-generator).
Why is a separate implementation needed in kotlin-sdk?

alpeshvas:
> Sorry, I don’t fully understand the motivation. There is already [xemantic-ai-tool-schema](https://github.com/xemantic/xemantic-ai-tool-schema), which was discussed in #32 There is also [victools/jsonschema-generator](https://github.com/victools/jsonschema-generator). Why is a separate implementation needed in kotlin-sdk?

It basically provides better and more kotlin idiomatic way to define MCP tool.
Earlier, to define a MCP tool (https://modelcontextprotocol.io/docs/concepts/tools), there is a lot which you need to write, which can be derived from the kotlin method signature.
```
server.addTool(
    name = "get_forecast",
    description = """
        Get weather forecast for a specific latitude/longitude
    """.trimIndent(),
    inputSchema = Tool.Input(
        properties = buildJsonObject {
            putJsonObject("latitude") {
                put("type", "number")
            }
            putJsonObject("longitude") {
                put("type", "number")
            }
        },
        required = listOf("latitude", "longitude")
    )
)
```

Now, with the proposed changes, defining a MCP tool is simpler with annotation ( uses kotlin annotations https://kotlinlang.org/docs/annotations.html) similar to how it's there in python sdk which uses decorators (similar concept to kotlin annotations).

check out this for example which we built using jar from this changes:

```
@McpTool(
    name = "get_forecast",
    description = "Get weather forecast for a specific latitude/longitude",
)
fun getForecastTool(latitude: Float, longitude: Float) 
```

so in short, it's less of a schema generators and more around simplifying MCP tool definitions for kotlin-sdk users.

devcrocod:
That’s understandable, but it doesn’t explain why not support one of the libraries I sent earlier.

Besides, the solution using annotations seems excessive to me at this point. 
It would be possible to support required without passing it as a parameter in the annotation, but instead by using the information about the default values of the parameters

Also, adding such a solution complicates support for multiplatform #70 , since reflection is platform-dependent

alpeshvas:
>  but it doesn’t explain why not support one of the libraries I sent earlier.

because they're MCP agnostic and have a generic purpose (schema generation)

> It would be possible to support required without passing it as a parameter in the annotation, but instead by using the information about the default values of the parameters


required is only one of the usecase, larger purpose is to reduce verbosity defining MCP tool. i.e
```
server.addTool(
    name = "kljlkjlkj",
    description = """kljkljjlk""".trimIndent(),
    inputSchema = Tool.Input(
        properties = buildJsonObject {
            putJsonObject("latitude") {
                put("type", "number")
            }
            putJsonObject("longitude") {
                put("type", "number")
            }
        },
        required = listOf("latitude", "longitude")
    )
)
```
currently, it's too verbose to define a tool and can be simplified.


> Also, adding such a solution complicates support for multiplatform https://github.com/modelcontextprotocol/kotlin-sdk/issues/70 , since reflection is platform-dependent

fair point, hence this PR adds support for doing it with annotations as an another way instead of replacing current one.

Stream29:
I'm working on a solution based on KSP.

With that, we can do this:

```
@McpServerComponent
public class MyServerComponent {
    @McpTool
    public suspend fun response(message: String): String {
        return "Hello, $message"
    }
}
```

Generate code from below code (@McpServerCompnent @McpTool annotation) with KSP as follow:

```
public fun MyServerComponent.tools(adapter: ServerAdapter = ServerAdapter()): List<RegisteredTool>= ...(Implementation)
```

And we can use it like:

```
val server = Server(...)
val component = MyServerComponent()
server.addTools(component.tools())
```

I think this plan is friendly for dependency injection and Kotlin Multiplatform, with less runtime cost.

build.gradle.kts
```diff
@@ -196,7 +197,9 @@ kotlin {
 
     explicitApi = ExplicitApiMode.Strict
 
-    jvmToolchain(21)
+    jvmToolchain {
+        languageVersion = JavaLanguageVersion.of(17) // Downgrade to JDK 17 which is more likely to be available
```
e5l:
```suggestion
        languageVersion = JavaLanguageVersion.of(17)
```

e5l (changes requested):
Hey @alpeshvas, thank you for the PR.

I'm fine with a new concept. It looks like it will be better to keep it separate from the core module:  the core will have raw implementation without kotlin-reflect dependencies, and annotations can be in a separate module. 

Could you please create a new module and extract the annotation and reflection logic?

ConstasJ:
I agree with @Stream29. I think this simple function doesn't need reflect, which is such a heavy tool. KSP is much simpler, and we (@Stream29 and me) have worked out a solution. Maybe we'll make another KSP-based solution later.

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -539,12 +539,12 @@ public open class Server(
         )
     }
 
-    private suspend fun handleListTools(): ListToolsResult {
+    public suspend fun handleListTools(): ListToolsResult {
```
alpeshvas, line 542: visibility change is needed to be accessible by the annotation module.

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -539,12 +539,12 @@ public open class Server(
...
         val toolList = tools.values.map { it.tool }
         return ListToolsResult(tools = toolList, nextCursor = null)
     }
 
-    private suspend fun handleCallTool(request: CallToolRequest): CallToolResult {
+    public suspend fun handleCallTool(request: CallToolRequest): CallToolResult {
```
alpeshvas, line 547: same as above.

alpeshvas:
> Hey @alpeshvas, thank you for the PR.
> 
> I'm fine with a new concept. It looks like it will be better to keep it separate from the core module: the core will have raw implementation without kotlin-reflect dependencies, and annotations can be in a separate module.
> 
> Could you please create a new module and extract the annotation and reflection logic?

Done.

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -105,7 +105,7 @@ public open class Server(
     /**
      * Registers a callback to be invoked when the server has completed initialization.
      */
-    public fun onInitalized(block: () -> Unit) {
+    public fun onInitialized(block: () -> Unit) {
```
alpeshvas, line 108: small typo

alpeshvas:
@e5l to run test  - ` ./gradlew :feature.annotation:test`

morisil:
@alpeshvas I see that the solution is reflection based, which would make it KMP incompatible. Please take a look how I approached it in this fully KMP-compatible `xemantic-ai-tool-schema` library:

https://github.com/xemantic/xemantic-ai-tool-schema/blob/main/src/commonMain/kotlin/generator/JsonSchemaGenerator.kt

It's `kotlinx.serialization` and meta-annotation based, probably KSP solution mentioned in another comment by @Stream29 would also keep it KMP-friendly. However by principle every LLM I know requires tool use / function calling input schema to be defined as a JSON Schema object, therefore implicit rule of inferring the schema from the sequence of function parameters sounds conceptually wrong. Tool input schema might be a very complex tree structure (which I am using quite often, even though due to misunderstanding behind the of original TypeScript MCP schema, it is not possible to express it in `kotlin-sdk`).

For the reference. Here is how it works in my KMP `anthropic-sdk-kotlin`:

```kotlin
@SerialName("get_weather")
@Description("Get the weather for a specific location")
data class GetWeather(val location: String)

fun main() = runBlocking {
    val tool = Tool<GetWeather> { "The weather is 73f" }
    val myTools = listOf(tool)
    val anthropic = Anthropic()

    val conversation = mutableListOf<Message>()
    conversation += "What is the weather in SF?"

    val initialResponse = client.messages.create {
        messages = conversation
        tools = myTools
    }
    println("Initial response: ${initialResponse.text}")

    conversation += initialResponse
    conversation += initialResponse.useTools()

    val finalResponse = client.messages.create {
        messages = conversation
        tools = myTools
    }
    println("Final response: ${finalResponse.text}")
}
```

https://github.com/xemantic/anthropic-sdk-kotlin#using-tools

This is the friendliness/compatibility level I wish to have in `kotlin-sdk` as well. This can be achieved by bridging with my `xemantic-ai-tool-schema-mdc` or even better, embracing the whole `xemantic-ai-tool-schema` under the umbrella of JetBrains, so it can be used directly in `kotlin-sdk` for MCP (unless JetBrains would accept external dependency on the `JsonSchema` object from `xemantic-ai-tool-schema` for the `inputSchema` type?

alpeshvas:
> @alpeshvas I see that the solution is reflection based, which would make it KMP incompatible. Please take a look how I approached it in this fully KMP-compatible `xemantic-ai-tool-schema` library:
> 
> 
> 
> https://github.com/xemantic/xemantic-ai-tool-schema/blob/main/src/commonMain/kotlin/generator/JsonSchemaGenerator.kt
> 
> 
> 
> It's `kotlinx.serialization` and meta-annotation based, probably KSP solution mentioned in another comment by @Stream29 would also keep it KMP-friendly. However by principle every LLM I know requires tool use / function calling input schema to be defined as a JSON Schema object, therefore implicit rule of inferring the schema from the sequence of function parameters sounds conceptually wrong. Tool input schema might be a very complex tree structure (which I am using quite often, even though due to misunderstanding behind the of original TypeScript MCP schema, it is not possible to express it in `kotlin-sdk`).
> 
> 
> 
> For the reference. Here is how it works in my KMP `anthropic-sdk-kotlin`:
> 
> 
> 
> ```kotlin
> 
> @SerialName("get_weather")
> 
> @Description("Get the weather for a specific location")
> 
> data class GetWeather(val location: String)
> 
> 
> 
> fun main() = runBlocking {
> 
>     val tool = Tool<GetWeather> { "The weather is 73f" }
> 
>     val myTools = listOf(tool)
> 
>     val anthropic = Anthropic()
> 
> 
> 
>     val conversation = mutableListOf<Message>()
> 
>     conversation += "What is the weather in SF?"
> 
> 
> 
>     val initialResponse = client.messages.create {
> 
>         messages = conversation
> 
>         tools = myTools
> 
>     }
> 
>     println("Initial response: ${initialResponse.text}")
> 
> 
> 
>     conversation += initialResponse
> 
>     conversation += initialResponse.useTools()
> 
> 
> 
>     val finalResponse = client.messages.create {
> 
>         messages = conversation
> 
>         tools = myTools
> 
>     }
> 
>     println("Final response: ${finalResponse.text}")
> 
> }
> 
> ```
> 
> 
> 
> https://github.com/xemantic/anthropic-sdk-kotlin#using-tools
> 
> 
> 
> This is the friendliness/compatibility level I wish to have in `kotlin-sdk` as well. This can be achieved by bridging with my `xemantic-ai-tool-schema-mdc` or even better, embracing the whole `xemantic-ai-tool-schema` under the umbrella of JetBrains, so it can be used directly in `kotlin-sdk` for MCP (unless JetBrains would accept external dependency on the `JsonSchema` object from `xemantic-ai-tool-schema` for the `inputSchema` type?

Fair points,that's the reason relfect based annotation has been moved to a different module, and can be iterated with KSP support or anything else in future iterations with separate PRs.

devcrocod:
the codebase has evolved significantly since this PR was opened, and the requested changes were not addressed.
I'm going to close this for now, feel free to reopen with an updated implementation if you'd like to continue working on this

PR #85, opened 2025-04-24
Add removing APIs for tools, prompts and resources

134130:
## Motivation and Context

- Add support for removing tools, prompts and resources.
- Currently, the SDK only supports adding APIs.

## How Has This Been Tested?

- With Unit Tests.

## Breaking Changes

- No.

## Types of changes

- [x] Documentation update

## Checklist


e5l (changes requested):
Hey @134130, thanks for the PR.

Could you tell me the reason for the change? I'm not sure I understand the use case

134130:
Hi @e5l, sorry for my lack of description.

The MCP is designed on the user's context. That is the reason why the client supports root API. For example:
- The server can provide directories based on user's root context

Also, there is a tool-list changed notifications channel. Which means the MCP Server is designed as the tools(resources, prompts) can be changed anytime.

So, there is no reason to restrict server providers from removing the tool, and that makes it difficult for me to make the rich MCP API for the users.

e5l:
That's a valid use case. If you want this use case to be fully supported, please consider writing a separate PR with documentation of it

PR #92, opened 2025-04-30
Disable configuration cache to fix jreleaser issue

e5l:


Mr3zee (approved):

StefMa:
Instead of removing it completely, you could just run the "release task" with `--no-configuration-cachd`

PR #99, opened 2025-05-05
fix(client) add encodeDefault for field with non spec default value

shiqicao:
`type` in `Tool.Input` has a default value, but the spec doesn't treat it as default value, therefore serailizer  should always write it. 

## Motivation and Context
Serialized `Tool` can be used directly in calling anthropic api (without SDK).

## How Has This Been Tested?
Passing all existing tests

## Breaking Changes
It is a breaking change if code depends on "missing" `type` behavior

## Types of changes
- [ ] Documentation update

## Checklist


e5l (changes requested):
Hey @shiqicao! Thanks for the PR. Could you add a unit test so I can merge?

e5l (approved):

PR #108, opened 2025-05-31
Remove fixed suffix /sse

adamglin0:
Remove the hardcoded /sse string in the sse client transport, which forcibly adds /sse to the path, making it very crude. 

When using all sse mcp servers, it's almost always in the format like `http://localhost:3000/sse`, and the forced addition of /sse leads to errors. In (npx @modelcontextprotocol/inspector), users also need to manually add the /sse suffix.

## Breaking Changes

If this PR is merged, consumers must manually add the /sse suffix to their URL.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt
+++ b/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt
@@ -68,7 +68,7 @@ public class SseClientTransport(
 
         session = urlString?.let {
             client.sseSession(
-                urlString = "$it/sse",
+                urlString = it,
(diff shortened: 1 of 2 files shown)


StefMa:
This is an breaking change!
Consumers might relay on /sse suffix.
They can easily fix it, but they have to be aware of it.

Mervap:
This is indeed a breaking change, but the current assumption simply doesn’t work for some MCP implementations.
A major version bump would clearly indicate that the release may include breaking changes

Please approve this PR, as it’s currently nearly impossible to use the client with servers that don’t end with `/sse`

devcrocod (dismissed):
lgtm
I’m considering adding /sse handling in separate pr

PR #87, opened 2025-04-27
Add Streamable Http Transport

SeanChinJunKai:
Closes #79 

Pending:
- Client Implementation of StreamableHttp
- Unit Tests
- Integration Tests

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #112, opened 2025-06-02
Fix StreamableHttpServerTransport review comments

michaellatman:
## Summary

This PR addresses all review comments from PR #87:

- Added comprehensive KDoc documentation for all public APIs
- Improved Accept header validation to use proper matching instead of contains
- Fixed ContentType header key to use HttpHeaders.ContentType
- Corrected parseBody error message for invalid JSON format
- Added unit tests for StreamableHttpServerTransport
- Did NOT add callMapping.clear() in close() method after checking TypeScript SDK

### Documentation
- Added KDoc comments for the class constructor parameters
- Added KDoc comments for the sessionId property
- Added detailed KDoc comments for all public methods

### Header Validation
- Changed Accept header validation to properly parse and match media types
- Now correctly handles headers with parameters (e.g., text/event-stream;q=0.9)

### Bug Fixes
- Fixed ContentType header to use HttpHeaders.ContentType instead of ContentType.toString()
- Fixed error message in parseBody method from "Server already initialized" to "Body must be a JSON object or array"

### CallMapping Handling
- After reviewing the TypeScript SDK implementation, I determined that callMapping should NOT be cleared in close()
- The entries are already properly cleaned up after each response is sent ([line 108](https://github.com/modelcontextprotocol/kotlin-sdk/blob/dc29d507ee2f276c48f746f02da86bdfdae32904/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt#L108))
- This matches the TypeScript implementation pattern

### Tests
- Added unit tests covering basic functionality (but more probably wanted)
- Tests verify proper initialization, start/close behavior, and configuration options

## Disclaimer

**Note**: I have not yet tried this library with a full application - I just wanted to help kick the stale review from PR #87. I can come back with more thorough testing around June 12 if you need additional help.

## Test Results
All tests are passing successfully.

### DIFF
--- a/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
+++ b/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
@@ -24,6 +24,11 @@ import kotlin.uuid.Uuid
  * Server transport for StreamableHttp: this allows server to respond to GET, POST and DELETE requests. Server can optionally make use of Server-Sent Events (SSE) to stream multiple server messages.
  *
  * Creates a new StreamableHttp server transport.
+ * 
+ * @param isStateful If true, the server will generate and maintain session IDs for each client connection.
+ *                   Session IDs are included in response headers and must be provided by clients in subsequent requests.
+ * @param enableJSONResponse If true, the server will return JSON responses instead of starting an SSE stream.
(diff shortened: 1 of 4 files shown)


michaellatman:
Definitely, I’ll need to revisit this. If no one else gets to it, I’ll probably come work on the client project as well.

devcrocod:
Obsolete

Issue #114, opened 2025-06-02
kotlin-mcp-server example project fails by not being able to find the wasm target.


Bwvolleyball:
**Describe the bug**
I'm trying to build the example `kotlin-mcp-server` project because I want to use the wasm target, but it fails because it can't find a wasm target of the kotlin mcp sdk.

**To Reproduce**

1. Go to the `kotlin-mcp-server` example project on the `main` branch, and run `./gradlew build`

**Expected behavior**
A build completes, giving me a wasm binary and a jvm binary of the example MCP server.

**Logs**

```
      > No matching variant of io.modelcontextprotocol:kotlin-sdk:0.5.0 was found. The consumer was configured to find a library for use during 'kotlin-api', preferably optimized for non-jvm, as well as attribute 'org.jetbrains.kotlin.klib.packaging' with value 'non-packed', attribute 'org.jetbrains.kotlin.platform.type' with value 'wasm', attribute 'org.jetbrains.kotlin.wasm.target' with value 'wasi' but:
          - Variant 'jvmApiElements-published' declares a library for use during compile-time, preferably optimized for standard JVMs:
              - Incompatible because this component declares a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'jvm' and the consumer needed a component, as well as attribute 'org.jetbrains.kotlin.platform.type' with value 'wasm'
```

**Additional context**

StefMa:
The samples using the published version of the mcp-kotlin library. The current version doesn't support wasm. Only the "dev" version on the main branch.

You can workaround it by switching to an composite build and use a project dependency instead. Should work I guess..

Maybe this should be even the default setup for the samples.

PR #119, opened 2025-06-23
Refactor samples

devcrocod:
#114

## Motivation and Context
- simplify the project structure
- test new features with existing samples

## How Has This Been Tested?
locally

## Breaking Changes
no

## Types of changes
- [ ] Documentation update

## Checklist


StefMa:
Instead of having one large build that includes the samples, you could also use an includeBuild aka [composite build](https://docs.gradle.org/current/userguide/composite_builds.html).
Just include the kotlin-mcp as an build in the samples and use it.
The benefit would be that both projects are somewhat standalone. Especially the main build.

e5l (changes requested):
This PR will make samples hard to use as standalone projects and complicate the project structure. Could you tell me why we include them into build?

StefMa:
> Could you tell me why we include them into build?

The rationale behind it is to test the current "development branch" / latest changes with the samples.
The problem is:
You have a project that points to the current development branch (`main` branch).
This includes the README and all the things.
You might introduce something (or APi changes etc.), but it literally can't be tested by "users" with the samples.
The samples point always to the latest stable version.
This might be confusing for people.

So I do the benfit of doing that.
However, as stated in my previous comment, maybe a better solution is to use an composite builds.
With that, the core sdk is still standalone, the samples are a *bit bound* to the location of the core sdk.
Otherwise, they are also standalone. You only have to point to the location of the core sdk (which is by default `../`).

Does this make sense? 🤔

devcrocod:
Obsolete

PR #128, opened 2025-06-26
Move samples deps to version catalog

StefMa:
This PR moves the dependencies of the samples to a single source - the root version catalog

Its a solution between a fully integregation of the samples (#119) into the project, while maintaining a single file for all dependencies.
With that, dependabot (see #121) is also able to update dependencies from the samples.

## How Has This Been Tested?

**I haven't tested the samples ❗**
Because of the dependency bumps (might happen), it might be the case that they don't work anymore.
I just run `./gradlew help` in the root project and the samples to make sure that dependency resolution works correctly.

## Breaking Changes

No breaking changes for consumers of the library itself.

## Types of changes

- [ ] Documentation update

## Checklist


devcrocod:
Obsolete
see: #208

PR #136, opened 2025-07-04
feat: add streamable http client

zdtango:
## Summary
Added streamable HTTP client support for MCP connections due to [SSE deprecation](https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse-deprecated)

## Motivation and Context

SSE transport is deprecated and many remote MCP servers only support streamable HTTP transport. Currently, the SDK only supports MCP over SSE, making it incompatible with remote MCP servers that have moved to streamable HTTP. This change adds the necessary streamable HTTP client support to maintain compatibility with current MCP server implementations.

## How Has This Been Tested?

Tested with multiple remote MCP servers including GitHub, Atlassian, and other MCP implementations that use streamable HTTP transport. Verified successful connection establishment, tool listing, and tool calling.

## Breaking Changes

None. This change only adds a new interface alongside the existing SSE transport without modifying existing functionality.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This addition ensures the SDK can connect to modern MCP servers as the ecosystem transitions away from deprecated SSE transport to streamable HTTP. The new transport method is essential for compatibility with remote MCP servers that no longer support SSE.

### DIFF
--- a/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
+++ b/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
@@ -1,0 +1,179 @@
+package io.modelcontextprotocol.kotlin.sdk.client
+
+import io.github.oshai.kotlinlogging.KotlinLogging
+import io.ktor.client.HttpClient
+import io.ktor.client.plugins.sse.ClientSSESession
+import io.ktor.client.plugins.sse.sseSession
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.client.request.post
+import io.ktor.client.request.setBody
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
+import io.ktor.http.append
+import io.ktor.http.contentType
+import io.ktor.http.isSuccess
+import io.modelcontextprotocol.kotlin.sdk.JSONRPCMessage
+import io.modelcontextprotocol.kotlin.sdk.shared.AbstractTransport
+import io.modelcontextprotocol.kotlin.sdk.shared.McpJson
+import kotlinx.coroutines.CoroutineName
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.launch
+import kotlin.concurrent.atomics.AtomicBoolean
+import kotlin.concurrent.atomics.ExperimentalAtomicApi
+
+private val logger = KotlinLogging.logger {}
+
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -0,0 +1,179 @@
...
+                            _onMessage(responseMessage)
+                        } catch (e: Exception) {
+                            logger.error(e) { "Error processing JSON response" }
+                            _onError(e)
+                        }
+                    }
+                }
+
+                ContentType.Text.EventStream.contentType -> {
+                    logger.trace { "Client received SSE stream in POST response. Messages will be handled by the main SSE session." }
```
AdrianMiska, line 149: I hope you don't mind me commenting, and correct me if I'm wrong, but I don't think this is correct. 

[Reading the relevant part of the spec,](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server) it seems to me that when the server returns `Content-Type: text/event-stream`, it's opening a new SSE connection for this request only, separate to the possibly established 'main' one that's kept open for notifications. In fact, the server _MUST NOT_ send responses on this main SSE session unless resuming after a disconnect.

This in turn means you'll need to process the SSE events right here.

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -0,0 +1,179 @@
...
+    private val scope by lazy { CoroutineScope(SupervisorJob()) }
+    private var sseJob: Job? = null
+    private var sessionId: String? = null
+
+    override suspend fun start() {
+        if (!initialized.compareAndSet(expectedValue = false, newValue = true)) {
+            error("StreamableHttpClientTransport already started!")
+        }
+        logger.debug { "Client transport starting..." }
+        startSseSession()
```
AdrianMiska, line 55: I also don't think you should start an SSE session here, as this will be called by the Client/Protocol during `connect` — before initialization is complete. Servers may not care, but this still [violates the spec](https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization), I think. This wasn't an issue in the SSE transport, as the SSE session is the medium by which the initialization is done, but that's not the case with HTTP transport.

The typescript-sdk handles this in the send method, [by checking whether the response is to the initialization request,](https://github.com/modelcontextprotocol/typescript-sdk/blob/11e84f0471b5f2dcc02d1751c17356d2370de6fe/src/client/streamableHttp.ts#L456) though I don't find that particularly elegant either.

alexandersumer:
Hello @zdtango , do you plan on adding unit tests?

devcrocod:
Based on this pr, I created a new one, with fixes and tests #147

devcrocod:
changes have been added in #147

PR #138, opened 2025-07-06
Add support for elicitation

### LINKED ISSUE #137: Support Elicitation

The 2025-06-18 version of the protocol introduces the "Elicitation" mechanism which allows servers to request additional information from users through the MCP client.

cf. https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation

- Add `elicitation` to `ClientCapabilities`
  - Handle `"elicitation/create"` in `Server.assertCapabilityForMethod()`
  - Handle `"elicitation/create"` in `Client.assertRequestHandlerCapability()`
- Add request and result types in `types.kt`
  - `ElicitationCreateRequest` server request
  - `ElicitationCreateResult` client result (with "accept"/"decline"/"cancel" actions)
- Handle `"elicitation/create"` requests in `Client`
- Add tests

N/A

**Additional context**

[MCP – Client Features – Elicitation](https://modelcontextprotocol.io/specification/2025-06-18/client/elicitation)

ptitjes:
Fixes #137

In this PR, I add:
- the request and result objects for the elicitation mechanism
- adds the client capabilities and capability checking on the client and server sides
- adds tests to `ClientTest`

## Motivation and Context

To support the new elicitation feature of the 2025-06-18 protocol version.

## How Has This Been Tested?

I added two tests in `ClientTest`. The tests are not extensive yet as I only check:
- that the server throws when the client doesn't support elicitation
- that the client correctly handles a simple elicitation request

Maybe should I add some more tests? Please advise me.

## Breaking Changes

N/A

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

1. I was not sure whether to add the tests to `ClientTest` or `ServerTest`. (It seems to me that the tests should be divided by feature and not by client/server side, as most features involve both client and server, but I respected how it is done today.)

2. I was not sure whether to add a method to `Client` to register an elicitation handler (i.e. wrap the `setRequestHandler(...)` call). Please advise me on what you think is best. (If that is best to add a wrapper method, maybe this can be implemented as a follow-up?)

3. The `CreateElicitationRequest.RequestedSchema` class I introduced is duplicated from `Tool.Input`. Maybe that should be factored-out in a `JsonSchema` top-level class in `types.kt`. I was not sure you would approve that, so I did not factor it out. Please tell me if you'd prefer I factor that out. (The support for the new Tool output schema in the latest protocol version will also need the same type.)


devcrocod (dismissed):
Your PR looks very great!

> I was not sure whether to add the tests to ClientTest or ServerTest. (It seems to me that the tests should be divided by feature and not by client/server side, as most features involve both client and server, but I respected how it is done today.)

You did right. This separation will actually be helpful later on when I split sdk into server and client modules. At that point, might also revisit the tests. Additionally, we have an open issue for adding integration tests.

> I was not sure whether to add a method to Client to register an elicitation handler (i.e. wrap the setRequestHandler(...) call). Please advise me on what you think is best. (If that is best to add a wrapper method, maybe this can be implemented as a follow-up?)

I think adding a dedicated method for elicitation would be a good idea. It’s not critical for this PR, but if you include it here, that would be great.

> The CreateElicitationRequest.RequestedSchema class I introduced is duplicated from Tool.Input. Maybe that should be factored-out in a JsonSchema top-level class in types.kt. I was not sure you would approve that, so I did not factor it out. Please tell me if you'd prefer I factor that out. (The support for the new Tool output schema in the latest protocol version will also need the same type.)

For now, your approach looks like a good solution. I’m not sure how the spec will evolve, and having separate classes might actually help us adapt to future changes more easily. Even if it means some duplication for the time being. I think you can leave it as it is for now and consider refactoring it in a separate PR

ptitjes:
@devcrocod Thanks a lot for your review.

> I think adding a dedicated method for elicitation would be a good idea. It’s not critical for this PR, but if you include it here, that would be great.

OK, I can add that in this PR. Would an `setElicitationHandler` method of the following shape be good for you?

```kotlin
fun Client.setElicitationHandler(handler: (CreateElicitationRequest) -> CreateElicitationResult) { /* ... */ }
```

> I’m not sure how the spec will evolve, and having separate classes might actually help us adapt to future changes more easily.

That makes sense!

ptitjes:
Added the new method and rebased on main.

devcrocod (approved):
lgtm

PR #172, opened 2025-07-18
Update CI workflows configuration for debugging

devcrocod:
## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/.github/workflows/gradle-publish.yml
+++ b/.github/workflows/gradle-publish.yml
@@ -70,14 +70,16 @@ jobs:
         run: |
           set -euo pipefail
           
-          if [[ ! -f build/jreleaser/output.properties ]]; then
-            echo "::error title=File Missing::The file 'build/jreleaser/output.properties' does not exist."
+          OUTPUT_PROPS="build/jreleaser/output.properties"
+          
+          if [[ ! -f "$OUTPUT_PROPS" ]]; then
+            echo "::error title=File Missing::The file '$OUTPUT_PROPS' does not exist."
             exit 1
           fi
           
-          DEPLOY_ID=$(grep '^deploymentId=' build/jreleaser/output.properties | cut -d= -f2 || true)
+          DEPLOY_ID=$(grep '^deploymentId=' "$OUTPUT_PROPS" | cut -d= -f2 || true)
           if [[ -z "$DEPLOY_ID" ]]; then
-            echo "::error title=Missing Deployment ID::The 'deploymentId' key was not found in 'build/jreleaser/output.properties'."
+            echo "::error title=Missing Deployment ID::The 'deploymentId' key was not found in '$OUTPUT_PROPS'."
             exit 1
           fi
           
@@ -86,7 +88,7 @@ jobs:
           AUTH=$(printf "%s:%s" "$OSSRH_USERNAME" "$OSSRH_TOKEN" | base64)
           echo "::add-mask::$AUTH"
           
-          STATUS_JSON=$(curl -s \
+          STATUS_JSON=$(curl -sS \
             -H "Authorization: Bearer $AUTH" \
             -H "Content-Type: application/json" \
             -X POST \
--- a/.github/workflows/gradle-publish.yml
+++ b/.github/workflows/gradle-publish.yml
@@ -9,7 +9,7 @@ name: Release
 
 on:
   release:
-    types: [created]
+    types: [ created ]
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -40,12 +40,12 @@ publishing {
 
 jreleaser {
     gitRootSearch = true
-    strict.set(true)
+    strict= true
```
sdubov: minor [styling]: missing space between `strict` and `=`.

.github/workflows/gradle-publish.yml
```diff
@@ -53,12 +50,68 @@ jobs:
           JRELEASER_GPG_SECRET_KEY: ${{ secrets.GPG_SECRET_KEY }}
           JRELEASER_GPG_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
 
-      - name: Publish to OSSRH repository
-        run: ./gradlew publish jreleaserFullRelease
+      - name: Publish to Maven Central Portal
+        id: publish
+        run: ./gradlew publish jreleaserFullRelease --info --stacktrace -Djreleaser.verbose=true
```
sdubov, line 55: Do we still need verbose level enabled here or was it just for debugging?

.github/workflows/gradle-publish.yml
```diff
@@ -53,12 +50,68 @@ jobs:
...
+      - name: Query Central Portal for validation errors
+        if: failure()
+        shell: bash
+        env:
+          OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
+          OSSRH_TOKEN: ${{ secrets.OSSRH_TOKEN }}
+        run: |
+          set -euo pipefail
+          
+          if [[ ! -f build/jreleaser/output.properties ]]; then
```
sdubov: Maybe, worth to extract the path to the file into a variable since you reuse it multiple times below.

.github/workflows/gradle-publish.yml
```diff
@@ -53,12 +50,68 @@ jobs:
...
+          
+          DEPLOY_ID=$(grep '^deploymentId=' build/jreleaser/output.properties | cut -d= -f2 || true)
+          if [[ -z "$DEPLOY_ID" ]]; then
+            echo "::error title=Missing Deployment ID::The 'deploymentId' key was not found in 'build/jreleaser/output.properties'."
+            exit 1
+          fi
+          
+          echo "Portal deploymentId: $DEPLOY_ID"
+          
+          AUTH=$(printf "%s:%s" "$OSSRH_USERNAME" "$OSSRH_TOKEN" | base64)
```
sdubov, line 98: Maybe, check that variables set before running the script?

.github/workflows/gradle-publish.yml
```diff
@@ -53,12 +50,68 @@ jobs:
...
+          echo "::add-mask::$AUTH"
```
sdubov, line 99: You add mask to the AUTH string with token here, but you print it without mask below. Why do we need to print the `$AUTH` string here and below at all?

sdubov (changes requested):
I have one small question about printing the `$AUTH` variable. Other than that, LGTM.

.github/workflows/gradle-publish.yml
devcrocod: It's necessary for debugging with the current release.
I'll remove it after resolving the publishing issue

.github/workflows/gradle-publish.yml
devcrocod: What do you mean?

.github/workflows/gradle-publish.yml
devcrocod: I'm not priting `$AUTH`, A mask will be used here, as I understand it, ensures security :

> When you mask a value, it is treated as a secret and will be redacted on the runner. For example, after you mask a value, you won’t be able to set that value as an output.

Also, I’m not printing auth later either, since I’m using `-s` with `curl`

.github/workflows/gradle-publish.yml
sdubov:
```
if [ -z "$OSSRH_USERNAME" ]; then
    echo "::error title='OSSRH_USERNAME' is not set."
    exit 1
 fi

if [ -z "$OSSRH_TOKEN" ]; then
    echo "::error title='OSSRH_TOKEN' is not set."
    exit 1
 fi
```
Please ignore, if we are sure that variables exists at that point.

devcrocod: okay,
make sense

.github/workflows/gradle-publish.yml
sdubov: Thank you. I've noticed now.

sdubov (approved):

tiginamaria (approved):

PR #198, opened 2025-07-25
Introduce server sessions

tiginamaria:
Introduce a `ServerSession` to manage multiple connections to the mcp server. `Protocall` logic of the `Server` was moved to  `ServerSession`,  while MCP resources logic remained in `Server`

## Motivation and Context

1. Fix the multiple connection issues reported for sse transport
2. Avoid server duplications during configurations, now every connection creates a session, but the server stays single

## How Has This Been Tested?

Added `SseIntegrationTest` and `WebSocketIntegrationTest` integration tests to check connection with single and multiple clients

## Breaking Changes

This change caused several depreciations:
1. The `Server` -> `Protocol` inheritance (which caused single connection issue) should be removed in future releases, now, all the methods inherited from `Protocall` marked as `Deprecated` with a `Warning`
2. Websocket ktor server initialization changes in order not to create server instance per connection

## Types of changes

- [ ] Documentation update

## Checklist


src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/WebSocketMcpKtorServerExtensions.kt
```diff
@@ -1,19 +1,105 @@
...
+internal suspend fun WebSocketServerSession.mcpWebSocketEndpoint(
+    block: () -> Server
+) {
+    logger.info { "Ktor Server establishing new connection" }
+    val transport = createMcpTransport(this)
+    val server = block()
+    var session: ServerSession? = null
+    try {
+        session = server.connectSession(transport)
+        awaitCancellation()
```
tiginamaria, line 73: Finishes without `await` in tests, not sure if it's a right approach to fix it, please take a look here! Really need some help here

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -156,9 +130,104 @@ public open class Server(
...
+
+    /**
+     * Starts a new server session with the given transport and initializes
+     * internal request handlers based on the server's capabilities.
+     *
+     * @param transport The transport layer to connect the session with.
+     * @return The initialized and connected server session.
+     */
+    // TODO: Rename connectSession to connect after the full connect deprecation
+    public suspend fun connectSession(transport: Transport): ServerSession {
```
tiginamaria: It should be renamed to connect after several deprecation cycles, but now I can not name it like this due to `Protocal` inheritance

api/kotlin-sdk.api
```diff
@@ -2934,6 +2934,11 @@ public final class io/modelcontextprotocol/kotlin/sdk/client/KtorClientKt {
 	public static synthetic fun mcpSseTransport-5_5nbZA$default (Lio/ktor/client/HttpClient;Ljava/lang/String;Lkotlin/time/Duration;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/modelcontextprotocol/kotlin/sdk/client/SseClientTransport;
 }
 
+public final class io/modelcontextprotocol/kotlin/sdk/client/MainKt {
```
devcrocod: It seems that `main.kt` was used for local testing and was mistakenly included in the API. Please remove it

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -8,94 +8,90 @@ import io.ktor.server.response.respond
...
     }
 
     post {
-        mcpPostEndpoint(transports)
+        mcpPostEndpoint(sseTransportManager)
     }
 }
 
-@Suppress("FunctionName")
-@Deprecated("Use mcp() instead", ReplaceWith("mcp(block)"), DeprecationLevel.WARNING)
```
devcrocod, line 45: I don’t think the deprecated functions should have been removed
instead, the `DeprecationLevel` should be increased to `Error`

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -83,23 +85,35 @@ public open class Server(
     private val serverInfo: Implementation,
     options: ServerOptions,
 ) : Protocol(options) {
+    private val sessions = atomic(persistentSetOf<ServerSession>())
```
devcrocod: Why is `set` instead of `queue`?

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -111,8 +125,9 @@ public open class Server(
         get() = _resources.value
 
     init {
-        logger.debug { "Initializing MCP server with capabilities: $capabilities" }
+        logger.debug { "Initializing MCP server with option: $options" }
```
devcrocod: `toString` is not overridden for `ServerOptions`, so this will output information that is useless for the user

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -156,9 +171,104 @@ public open class Server(
         }
     }
 
+    @Deprecated(
+        "Will be removed with Protocol inheritance. Use connectSession instead.",
```
devcrocod: > `Use connectSession instead.` 

The deprecation message doesn’t seem entirely accurate. `connectionSession` is a function and it doesn’t replace `onClose` function.
I think it would be better to simply state that the architecture has changed to use Server Sessions

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
devcrocod: Maybe we should still consider changing the behavior of `connect`?

If we simply mark `connect` as deprecated, we’ll eventually have to remove it. But then we might end up deprecating `connectSession` and reintroducing `connect` again. And that will create new issues for us, because we forced users to move from `connect` to `connectSession`.

In that case, I think we have two options:
- Change the current `connect` (which would be a breaking change)
- Deprecate `connect` and stick with `connectSession`

In my opinion, the first option is preferable

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -667,6 +806,33 @@ public open class Server(
         return ListResourceTemplatesResult(listOf())
     }
 
+
+    @Deprecated(
```
devcrocod: I’m a bit confused about this part.

First, we can remove private functions, they’re not part of the public API.

Second, this is used in the `init` block. 
What will the behavior be in that case? Should we consider rewriting the init block as well?

devcrocod (changes requested):
The PR looks good to me overall!

Please take a look at my comments.
My main concern is about the behavior of the `Server`
It would also be great to update the existing samples and the README accordingly

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/Client.kt
```diff
@@ -158,14 +158,13 @@ public open class Client(
             serverVersion = result.serverInfo
 
             notification(InitializedNotification())
+        } catch (error: CancellationException) {
+            throw IllegalStateException("Error connecting to transport: ${error.message}")
```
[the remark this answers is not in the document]
e5l: please also consider including error as a cause

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -111,8 +125,9 @@ public open class Server(
         get() = _resources.value
 
     init {
-        logger.debug { "Initializing MCP server with capabilities: $capabilities" }
+        logger.debug { "Initializing MCP server with option: $options" }
 
+        // TODO: Remove all after Protocol inheritance
```
e5l: please log an issue for TODO and mention it in the comment (same for other TODOs)

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
```diff
@@ -0,0 +1,343 @@
...
+     * Asserts that the server can handle the specified request method.
+     *
+     * Throws [IllegalStateException] if the server does not have the capabilities required to handle this request.
+     *
+     * @param method The request method.
+     */
+    override fun assertRequestHandlerCapability(method: Method) {
+        logger.trace { "Asserting request handler capability for method: ${method.value}" }
+        when (method.value) {
+            "sampling/createMessage" -> {
```
e5l: Please consider extracting constants or reusing the existing

e5l (changes requested):
Hey @tiginamaria, thank you for the PR. Could you add a sample as a docs page illustrating what the difference will be between kotlin and the original TypeScript API?

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
tiginamaria: Totally agree, remove Protocol inheritance from the server, which cause this naming problem

src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
tiginamaria: List is fine here, yes

devcrocod (dismissed):
lgtm!

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -13,97 +13,106 @@ import io.ktor.server.routing.routing
...
 import io.ktor.utils.io.KtorDsl
+import kotlinx.atomicfu.AtomicRef
+import kotlinx.atomicfu.atomic
+import kotlinx.atomicfu.update
+import kotlinx.collections.immutable.PersistentMap
+import kotlinx.collections.immutable.toPersistentMap
 
 private val logger = KotlinLogging.logger {}
 
+internal class SseTransportManager(transports: Map<String, SseServerTransport> = emptyMap()) {
```
kpavlov, line 25: nit: using typealias for session ID could make code more readable and typesafe, e.g.:

```suggestion

private typealias SessionId = String

internal class SseTransportManager(transports: Map<SessionId, SseServerTransport> = emptyMap()) {
```

... and session ID type can be changed to `SessionId` in all other places

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+import kotlin.test.assertTrue
+import io.ktor.client.engine.cio.CIO as ClientCIO
+import io.ktor.client.plugins.websocket.WebSockets as ClientWebSocket
+import io.ktor.server.cio.CIO as ServerCIO
+import io.ktor.server.websocket.WebSockets as ServerWebSockets
+
+class WebSocketIntegrationTest {
+
+    @Test
+    fun `client should be able to connect to websocket server 2`() = runTest {
```
kpavlov: nit: a bit concise 
```suggestion
    fun `client should connect to websocket server`() = runTest(timeout = 5.seconds) {
```

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+        var server: EmbeddedServer<CIOApplicationEngine, CIOApplicationEngine.Configuration>? = null
+        var client: Client? = null
+
+        try {
+            withContext(Dispatchers.Default) {
+                withTimeout(1000) {
+                    server = initServer()
+                    val port = server.engine.resolvedConnectors().first().port
+                    client = initClient(serverPort = port)
```
kpavlov: let's actually verify something

```suggestion
                    client = initClient(serverPort = port)
                    
                    val serverVersion = client.serverVersion
                    assertEquals("websocket-server", serverVersion?.name)
                    assertEquals("1.0.0", serverVersion?.version)
```

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+class WebSocketIntegrationTest {
+
+    @Test
+    fun `client should be able to connect to websocket server 2`() = runTest {
+        var server: EmbeddedServer<CIOApplicationEngine, CIOApplicationEngine.Configuration>? = null
+        var client: Client? = null
+
+        try {
+            withContext(Dispatchers.Default) {
+                withTimeout(1000) {
```
kpavlov: Can it be replaced with `runTest(timeout=5.seconds){ ... }`?
```suggestion
```

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+                    server = initServer()
```
kpavlov: Since the server is the same, I would create the server instance once for test class (a.k.a. `@BeforeAll`). This might also help to catch concurrency problems since the server instance will be shared. This is valid for all integration tests

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+                    val port = server.engine.resolvedConnectors().first().port
+                    client = initClient(serverPort = port)
+                }
+            }
+        } finally {
+            client?.close()
+            server?.stop(1000, 2000)
```
kpavlov: If server is shared, then it's not needed, test verifies that all connections are closed and server is stopped. But it's a different test

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
```diff
@@ -0,0 +1,207 @@
...
+
+        try {
+            withContext(Dispatchers.Default) {
+                withTimeout(1000) {
+                    server = initServer()
+                    val port = server.engine.resolvedConnectors().first().port
+                    client = initClient("Client A", port)
+
+                    val promptA = getPrompt(client, "Client A")
+                    assertTrue { "Client A" in promptA }
```
kpavlov: nit: Why not try kotest-assertions, since the dependency is alread in the project?

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/SseIntegrationTest.kt
```diff
@@ -1,63 +1,206 @@
...
 import io.ktor.server.cio.CIO as ServerCIO
 import io.ktor.server.sse.SSE as ServerSSE
 
-private const val URL = "127.0.0.1"
-
 class SseIntegrationTest {
     @Test
     fun `client should be able to connect to sse server`() = runTest {
-        val serverEngine = initServer()
+        var server: EmbeddedServer<CIOApplicationEngine, CIOApplicationEngine.Configuration>? = null
```
kpavlov, line 36: Server instance can be reused, [like in another test](https://github.com/modelcontextprotocol/kotlin-sdk/pull/198/files#diff-b188c15b1d3b462026c5d820f61405d2f3e8887c8d247d580c24fc9bc8eb13cf). I would create a helper function for server initialization and reuse it across tests classes

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/SseIntegrationTest.kt
```diff
@@ -1,63 +1,206 @@
...
+        var client: Client? = null
+
+        try {
+            withContext(Dispatchers.Default) {
+                withTimeout(1000) {
```
kpavlov: same as [here](https://github.com/modelcontextprotocol/kotlin-sdk/pull/198/files#r2300488275)

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseTransportTest.kt
```diff
@@ -70,19 +70,6 @@ class SseTransportTest : BaseTransportTest() {
...
             routing {
                 mcp { mcpServer }
-//                sse {
-//                    mcpSseTransport("", transports).apply {
-//                        onMessage {
-//                            send(it)
-//                        }
-//
-//                        start()
-//                    }
```
kpavlov, line 80: Is it possible to remove commented code?

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/WebSocketMcpKtorClientExtensions.kt
```diff
@@ -36,6 +39,8 @@ public suspend fun HttpClient.mcpWebSocket(
             version = LIB_VERSION,
         ),
     )
+    logger.debug { "Client started to connect to server" }
```
kpavlov, line 42: nit: should it be trace level? It might be too verbose for debug.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/WebSocketMcpTransport.kt
```diff
@@ -53,7 +58,8 @@ public abstract class WebSocketMcpTransport : AbstractTransport() {
             while (true) {
                 val message = try {
                     session.incoming.receive()
-                } catch (_: ClosedReceiveChannelException) {
+                } catch (e: ClosedReceiveChannelException) {
```
kpavlov, line 61: `e` is actually not used. Can be reverted to `_`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
```diff
@@ -0,0 +1,371 @@
...
+     */
+    public var clientVersion: Implementation? = null
+        private set
+
+    /**
+     * Registers a callback to be invoked when the server has completed initialization.
+     */
+    public fun onInitialized(block: () -> Unit) {
+        val old = _onInitialized
+        _onInitialized = {
```
kpavlov, line 75: maybe `_onInitialized` and `_onClose` should be changed to `List<() -> Unit>`. 

nit: Also, is it expected that all subsequent callback blocks will be executed after exception in the previous block? Should every block be wrapped with `try/catch`? Can/should callbacks be called in parallel? It's a design decision.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
```diff
@@ -0,0 +1,371 @@
...
+        _onClose()
+    }
+
+    /**
+     * Sends a ping request to the client to check connectivity.
+     *
+     * @return The result of the ping request.
+     * @throws IllegalStateException If for some reason the method is not supported or the connection is closed.
+     */
+    public suspend fun ping(): EmptyRequestResult = request<EmptyRequestResult>(PingRequest())
```
kpavlov, line 106: nit: all explicit type arguments can be removed
```suggestion
    public suspend fun ping(): EmptyRequestResult = request<>(PingRequest())
```

kpavlov (dismissed):
Thank you, @tiginamaria, @devcrocod. Callback chaining implementation is very simple and it can be improved in a separate PR along with the tests.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
devcrocod: In my opinion, it’s better to use a typealias for complex composite types
replacing a regular string doesn’t improve code readability and only adds extra cognitive complexity

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
devcrocod: Agreed, but why exactly 5 seconds?

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
devcrocod: it's commonTest

devcrocod:
@tiginamaria 
please check tests failed: 
```
io.modelcontextprotocol.kotlin.sdk.integration.SseIntegrationTest.client should be able to connect to sse server[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14

io.modelcontextprotocol.kotlin.sdk.integration.SseIntegrationTest.single sse connection[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14

io.modelcontextprotocol.kotlin.sdk.integration.SseIntegrationTest.multiple sse connections[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14

io.modelcontextprotocol.kotlin.sdk.integration.WebSocketIntegrationTest.client should be able to connect to websocket server 2[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14

io.modelcontextprotocol.kotlin.sdk.integration.WebSocketIntegrationTest.single websocket connection[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14

io.modelcontextprotocol.kotlin.sdk.integration.WebSocketIntegrationTest.multiple websocket connections[js, node] FAILED
    IllegalStateException at /mnt/agent/work/8d547b974a7be21f/ktor-server/ktor-server-cio/jsAndWasmShared/src/io/ktor/server/cio/internal/CoroutineUtils.jsAndWasmShared.kt:14
```
    
`runBlocking` is not supported for wasm

devcrocod:
I found the problem: these tests use `stop` for the Ktor server, but it should be `stopSuspend`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
tiginamaria: Can wait for another pr with callbacks refactorings

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/WebSocketIntegrationTest.kt
tiginamaria: But I need to free the port and the behavior you described is not expected, no?

kotlin-sdk-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractResourceIntegrationTest.kt
```diff
@@ -310,4 +307,16 @@ abstract class AbstractResourceIntegrationTest : KotlinTestBase() {
             assertTrue(result.contents.isNotEmpty(), "Result contents should not be empty")
         }
     }
+
```
tiginamaria: For me: remove

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -13,97 +13,106 @@ import io.ktor.server.routing.routing
...
+    private val transports: AtomicRef<PersistentMap<String, SseServerTransport>> = atomic(transports.toPersistentMap())
```
kpavlov, line 26: We have decided to use ConcurrentMaps from ktor, right?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -108,55 +76,83 @@ public open class Server(private val serverInfo: Implementation, options: Server
     public val resources: Map<String, RegisteredResource>
         get() = _resources.value
 
-    init {
-        logger.debug { "Initializing MCP server with capabilities: $capabilities" }
+    public suspend fun close() {
+        logger.debug { "Closing MCP server" }
+        sessions.value.forEach { it.close() }
```
kpavlov, line 81: Let's also clear sessions afterwards. Also, let's add try-catch around session.close() to make sure all resources are released. Let's also add unit test about it: verify that method close() is called for all sessions

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -77,26 +54,17 @@ public class ServerOptions(public val capabilities: ServerCapabilities, enforceS
  * @param serverInfo Information about this server implementation (name, version).
  * @param options Configuration options for the server.
  */
-public open class Server(private val serverInfo: Implementation, options: ServerOptions) : Protocol(options) {
+public open class Server(private val serverInfo: Implementation, private val options: ServerOptions) {
+    private val sessions = atomic(persistentListOf<ServerSession>())
```
kpavlov, line 58: When one session is closed, it should be removed from the collection. It might be better to use `ConcurrentMap<SessionId, ServerSession>`. But I failed to find where it is happening...

kotlin-sdk-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/SseIntegrationTest.kt
```diff
@@ -1,63 +1,195 @@
...
 
-private const val URL = "127.0.0.1"
-
 class SseIntegrationTest {
     @Test
-    fun `client should be able to connect to sse server`() = runTest {
-        val serverEngine = initServer()
+    fun `client should be able to connect to sse server`() = runTest(timeout = 5.seconds) {
+        var server: EmbeddedServer<CIOApplicationEngine, CIOApplicationEngine.Configuration>? = null
+        var client: Client? = null
```
kpavlov, line 37:
```suggestion
        lateinit var client: Client
```
?

e5l (approved):

kpavlov:
Merging with the #297 in mind

Issue #111, opened 2025-06-01
Support other engines (client/server)


MukjepScarlet:
We usually have an existing applicable engine(dependency) in the project, like Netty(server) and OkHttp(client), this can minimize the dependency list

PR #223, opened 2025-08-13
Remove dependency on CIO engine

MukjepScarlet:
Replace the CIO dependency with only server/client core dependency.

## Motivation and Context

In order to minimize the size of package. There are probably other existing engines like Netty/Tomcat(server) or Java/OkHttp/Apache(client) in the classpath. So we can reuse existing infrastructure.

## How Has This Been Tested?

I have run:
- kotlin-sdk-client: jvmTest, jsTest
- kotlin-sdk-test: jvmTest, jsTest
- kotlin-sdk-server: allTests
- All `build` of samples

I have an issue about `samples/kotlin-mcp-server`. `io.modelcontextprotocol` is not found in `commonMain`.

## Breaking Changes

Yes. They need to `implementation` CIO engine manually if they use it in their app.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

closes https://github.com/modelcontextprotocol/kotlin-sdk/issues/111

### DIFF
--- a/samples/kotlin-mcp-client/build.gradle.kts
+++ b/samples/kotlin-mcp-client/build.gradle.kts
@@ -15,7 +15,7 @@ dependencies {
     implementation(libs.mcp.kotlin)
     implementation(libs.ktor.client.cio)
     implementation(libs.anthropic.java)
-    runtimeOnly(libs.slf4j.simple)
+    implementation(libs.slf4j.simple)
 }
 
 tasks.test {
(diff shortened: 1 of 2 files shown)


devcrocod:
Thanks @MukjepScarlet for your pr!

Sorry, I’m having trouble understanding the motivation behind this PR
Is the package size really so large that it needs to be reduced, especially considering that the user will still need a dependency on some engine anyway?

Also, the user can already choose any other engine. CIO is just the default engine

MukjepScarlet:
The main motivation is that the users of client applications like iOS/Android/Desktop would like to see a smaller package. And we usually want to integrate the lib into an existing application with some infrastructures, maybe tons of configurations, in this situation the lib-integrated engine is redundant.

kpavlov (commented):
Thank you, @MukjepScarlet 
It's totally make sense to remove KTor implementation from library dependencies, so users would not have to exclude CIO if they want to use another client or server. Let's mention it in the documentation/README too.

MukjepScarlet:
Waiting for test workflow

Issue #226, opened 2025-08-17
Client.connect with StreamableHttpClientTransport hangs indefinitely on invalid JSON response


AdrianMiska:
**Describe the bug**
When `McpJson.decodeFromString` fails inside `StreamableHttpClientTransport` `.send` or `.handleInlineSse`, the `Client`/`Protocol` is left waiting on a `CompletableDeferred` that never completes. As a result, the program hangs indefinitely with no indication of the underlying error.

**To Reproduce**
1. Create a `Client` 
2. Create a `StreamableHttpClientTransport`
3. Have the MCP server respond with invalid JSON (or trigger another deserialization error, e.g. same as in #189).
4. Call `client.connect(transport)`

**Observed behavior**
The program **hangs indefinitely** without surfacing the decoding failure.

**Expected behavior**
`client.connect` should fail by throwing an exception, making the error visible to the caller.

**Additional context**
What's happening is that `StreamableHttpClientTransport` passes the exception to its `onError` callback without re-throwing it. Meanwhile, `Protocol` is still awaiting completion of its `CompletableDeferred`, which only resolves when an actual message is received. Since that never happens, the `await` call blocks forever.

**Test to reproduce**
```kotlin
@Test
fun testClientConnectWithInvalidJson() = runTest {
    // Transport under test: respond with invalid JSON for the initialize request
    val transport = createTransport { _ ->
        respond(
            "this is not valid json",
            status = HttpStatusCode.OK,
            headers = headersOf(HttpHeaders.ContentType, ContentType.Application.Json.toString()),
        )
    }

    val client = Client(
        clientInfo = io.modelcontextprotocol.kotlin.sdk.Implementation(
            name = "test-client",
            version = "1.0",
        ),
    )

    try {
        // Real time-keeping is needed; otherwise Protocol will always throw TimeoutCancellationException in tests
        withContext(Dispatchers.Default.limitedParallelism(1)) {
            kotlinx.coroutines.withTimeout(5.seconds) {
                client.connect(transport)
            }
        }
        kotlin.test.fail("Expected client.connect to fail on invalid JSON response")
    } catch (e: TimeoutCancellationException) {
        // Current behavior: result.await() in Protocol is never resolved, leading to a hang.
        kotlin.test.fail("Client connect caused a hang", e)
    } catch (_: IllegalStateException) {
        // Expected behavior: connect finishes and fails with an exception.
    } catch (e: Throwable) {
[...]

PR #227, opened 2025-08-17
fix hang caused by StreamableHttpClientTransport (#226)

AdrianMiska:
Re-throw exceptions in `StreamableHttpClientTransport` so `Protocol` doesn't hang indefinitely.

## Motivation and Context
Fixes the issue described in #226

## How Has This Been Tested?
I've verified that my test case reproduces the error without my fix, and verified with breakpoints that the hang occurred where I described it.

## Breaking Changes
None

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
I am unsure about the dispatcher I used for the real-time timeout here. I could not get the test to work otherwise, as `kotlinx.coroutines.test` was always tripping my timeout, with or without the fix. This is seemingly unrelated to the test case though, as removing my timeout results in the [timeout in Protocol](https://github.com/modelcontextprotocol/kotlin-sdk/blob/6bae9874c0ba02c683dfe61a416eeee0de0c6f32/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt#L454) being tripped, even though that's not the point of the indefinite hang.

### DIFF
--- a/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransportTest.kt
+++ b/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransportTest.kt
@@ -417,6 +417,7 @@ class StreamableHttpClientTransportTest {
         }.onFailure { e ->
             when (e) {
                 is TimeoutCancellationException -> fail("Client connect caused a hang", e)
+
                 is IllegalStateException -> {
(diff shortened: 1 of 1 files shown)


AdrianMiska:
I just realized that at some point I removed the dispatcher I was talking about.

AdrianMiska:
I've just checked again that all tests on all platforms pass on my machine (except for browser, I couldn't get those to work). 
So this failure is either in the browser tests (seems unlikely to me if all other platforms pass) or something that only happens on the GitHub runner.

Issue #183, opened 2025-07-21
Server implementation of Streamable transport


devcrocod:

devcrocod:
#87

PR #235, opened 2025-08-25
Add Server Streamable Http Transport

### LINKED ISSUE #79: Adding Streaming HTTP Transport
It is part of specification upgrades https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

I would like to add the new transport

Hi @e5l and @devcrocod , could I go ahead with implementing this?

zshea:
Closes #79 #183 

Implement server StreamableHttpTransport (both stateful and stateless)
This is a workable solution, tested with Claude Code

Caveat:
1. The Ktor server SSE plugin only supports the GET method. Even if I hacked the routing to support the POST method, it behaves weirdly (see additional context). In this PR, the StreamableHttpTransport always works with `enableJsonResponse = true`. We can remove this limitation once the upstream has fixed this issue.
2. Because of the above issue, I can't write an integration test for this transport. The client `StreamHttpTransport` assumes the SSE stream while the server assumes the JSON response.
3. I didn't update the documentation, hope this PR won't have significant changes. I'll update the documentation after this PR is merged.

## Motivation and Context
It is painful to support `/sse` + `/message` endpoint on production. Hopefully we can close this issue with this PR.

## How Has This Been Tested?
It has been tested in my environment with Claude Code (1.0.89) MCP with both `mcpStreamableHttp` and `mcpStatelessStreamableHttp`

## Breaking Changes
No, everything is backward compatible

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
1. Ktor server SSE plugin with POST method
It is workable with the following implementation:
```Kotlin
route("/sse", HttpMethod.POST) {
    sse {
        call.response.status(HttpStatusCode.NotAcceptable)
        send(event = "message", data = "1")
    }
}
```
But the response is always 200, even if we set the status explicitly. And the connection won't be closed with a 4xx or 5xx response. It breaks the MCP spec and can't be used directly. I think we need to create issues to upstream.

2. Thanks @SeanChinJunKai and @devcrocod 's work on #87. I got some insights there and created this PR.

### DIFF
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
@@ -249,14 +249,14 @@ public data class JSONRPCNotification(
  */
 @Serializable
 public class JSONRPCResponse(
-    public val id: RequestId,
+    public val id: RequestId?,
     public val jsonrpc: String = JSONRPC_VERSION,
     public val result: RequestResult? = null,
     public val error: JSONRPCError? = null,
 ) : JSONRPCMessage {
 
     public fun copy(
-        id: RequestId = this.id,
+        id: RequestId? = this.id,
         jsonrpc: String = this.jsonrpc,
         result: RequestResult? = this.result,
         error: JSONRPCError? = this.error,
@@ -292,8 +292,12 @@ public sealed interface ErrorCode {
  * A response to a request that indicates an error occurred.
  */
 @Serializable
-public data class JSONRPCError(val code: ErrorCode, val message: String, val data: JsonObject = EmptyJsonObject) :
-    JSONRPCMessage
+public data class JSONRPCError(
+    val id: RequestId? = null,
+    val code: ErrorCode,
+    val message: String,
+    val data: JsonObject = EmptyJsonObject,
+) : JSONRPCMessage
 
 /**
  * Base interface for notification parameters with optional metadata.
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
@@ -4,6 +4,7 @@ import io.github.oshai.kotlinlogging.KotlinLogging
 import io.ktor.http.HttpStatusCode
 import io.ktor.server.application.Application
 import io.ktor.server.application.install
+import io.ktor.server.request.header
 import io.ktor.server.response.respond
 import io.ktor.server.routing.Routing
 import io.ktor.server.routing.RoutingContext
@@ -15,6 +16,7 @@ import io.ktor.server.sse.ServerSSESession
 import io.ktor.server.sse.sse
 import io.ktor.util.collections.ConcurrentMap
 import io.ktor.utils.io.KtorDsl
+import io.modelcontextprotocol.kotlin.sdk.ErrorCode
 
 private val logger = KotlinLogging.logger {}
 
@@ -64,6 +66,51 @@ public fun Application.mcp(block: ServerSSESession.() -> Server) {
     }
 }
 
+@KtorDsl
+public fun Application.mcpStreamableHttp(
+    enableDnsRebindingProtection: Boolean = false,
+    allowedHosts: List<String>? = null,
+    allowedOrigins: List<String>? = null,
+    eventStore: EventStore? = null,
+    block: RoutingContext.() -> Server,
+) {
+    val transports = ConcurrentMap<String, StreamableHttpServerTransport>()
+
+    routing {
+        post("/mcp") {
+            mcpStreamableHttpEndpoint(
+                transports,
+                enableDnsRebindingProtection,
+                allowedHosts,
+                allowedOrigins,
+                eventStore,
+                block,
+            )
+        }
+    }
+}
+
+@KtorDsl
+public fun Application.mcpStatelessStreamableHttp(
+    enableDnsRebindingProtection: Boolean = false,
+    allowedHosts: List<String>? = null,
+    allowedOrigins: List<String>? = null,
+    eventStore: EventStore? = null,
+    block: RoutingContext.() -> Server,
+) {
+    routing {
+        post("/mcp") {
+            mcpStatelessStreamableHttpEndpoint(
+                enableDnsRebindingProtection,
+                allowedHosts,
+                allowedOrigins,
+                eventStore,
+                block,
+            )
+        }
+    }
+}
+
 private suspend fun ServerSSESession.mcpSseEndpoint(
     postEndpoint: String,
     transports: ConcurrentMap<String, SseServerTransport>,
@@ -92,6 +139,88 @@ internal fun ServerSSESession.mcpSseTransport(
     logger.info { "New SSE connection established and stored with sessionId: ${transport.sessionId}" }
 
     return transport
+}
+
+private suspend fun RoutingContext.mcpStreamableHttpEndpoint(
+    transports: ConcurrentMap<String, StreamableHttpServerTransport>,
+    enableDnsRebindingProtection: Boolean = false,
+    allowedHosts: List<String>? = null,
+    allowedOrigins: List<String>? = null,
+    eventStore: EventStore? = null,
+    block: RoutingContext.() -> Server,
+) {
+    val sessionId = this.call.request.header(MCP_SESSION_ID_HEADER)
+    val transport = if (sessionId != null && transports.containsKey(sessionId)) {
+        transports[sessionId]!!
+    } else if (sessionId == null) {
+        val transport = StreamableHttpServerTransport(
+            enableDnsRebindingProtection = enableDnsRebindingProtection,
+            allowedHosts = allowedHosts,
+            allowedOrigins = allowedOrigins,
+            eventStore = eventStore,
+            enableJsonResponse = true,
+        )
+
+        transport.setOnSessionInitialized { sessionId ->
+            transports[sessionId] = transport
+
+            logger.info { "New StreamableHttp connection established and stored with sessionId: $sessionId" }
+        }
+
+        val server = block()
+        server.onClose {
+            logger.info { "Server connection closed for sessionId: ${transport.sessionId}" }
+        }
+
+        server.connect(transport)
+
+        transport
+    } else {
+        null
+    }
+
+    if (transport == null) {
+        this.call.reject(
+            HttpStatusCode.BadRequest,
+            ErrorCode.Unknown(-32000),
+            "Bad Request: No valid session ID provided",
+        )
+        return
+    }
+
+    transport.handleRequest(null, this.call)
+    logger.debug { "Server connected to transport for sessionId: ${transport.sessionId}" }
+}
+
+private suspend fun RoutingContext.mcpStatelessStreamableHttpEndpoint(
+    enableDnsRebindingProtection: Boolean = false,
+    allowedHosts: List<String>? = null,
+    allowedOrigins: List<String>? = null,
+    eventStore: EventStore? = null,
+    block: RoutingContext.() -> Server,
+) {
+    val transport = StreamableHttpServerTransport(
+        enableDnsRebindingProtection = enableDnsRebindingProtection,
+        allowedHosts = allowedHosts,
+        allowedOrigins = allowedOrigins,
+        eventStore = eventStore,
+        enableJsonResponse = true,
+    )
+    transport.setSessionIdGenerator(null)
+
+    logger.info { "New stateless StreamableHttp connection established without sessionId" }
+
+    val server = block()
+
+    server.onClose {
+        logger.info { "Server connection closed without sessionId" }
+    }
+
+    server.connect(transport)
+
+    transport.handleRequest(null, this.call)
+
+    logger.debug { "Server connected to transport without sessionId" }
 }
 
 internal suspend fun RoutingContext.mcpPostEndpoint(transports: ConcurrentMap<String, SseServerTransport>) {
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
@@ -1,0 +1,575 @@
+package io.modelcontextprotocol.kotlin.sdk.server
+
+import io.ktor.http.ContentType
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpMethod
+import io.ktor.http.HttpStatusCode
+import io.ktor.server.application.ApplicationCall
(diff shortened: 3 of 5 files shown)


zshea:
Hi @e5l @devcrocod @SeanChinJunKai , would you mind taking a look?

zshea:
Hi @devcrocod , would you mind taking a look when available?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -249,14 +249,14 @@ public data class JSONRPCNotification(
  */
 @Serializable
 public class JSONRPCResponse(
-    public val id: RequestId,
+    public val id: RequestId?,
```
kpavlov:
```suggestion
    public val id: RequestId? = null,
```

kpavlov: According to the [spec](https://modelcontextprotocol.io/specification/2025-06-18/schema#jsonrpcresponse), the `id` is required field

kpavlov (changes requested):
Thank you @zshea  for the PR
Would it be possible to add integration tests to make sure it actually works?
Also, `id` is mandatory for `JSONRPCRequest`, `JSONRPCResponse`, so let's keep it non-nullable

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -64,6 +66,51 @@ public fun Application.mcp(block: ServerSSESession.() -> Server) {
...
+                allowedHosts,
+                allowedOrigins,
+                eventStore,
+                block,
+            )
+        }
+    }
+}
+
+@KtorDsl
```
kpavlov, line 109: Missing KDoc

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -0,0 +1,575 @@
...
+        sessionId?.let { this.response.headers.append(MCP_SESSION_ID_HEADER, it) }
+        this.response.status(HttpStatusCode.OK)
+    }
+}
+
+internal suspend fun ApplicationCall.reject(status: HttpStatusCode, code: ErrorCode, message: String) {
+    this.response.status(status)
+    this.respond(
+        JSONRPCResponse(
+            id = null,
```
dvilker: Yom can use `RequestId.StringId("server-error")` as in the Python SDK

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -0,0 +1,575 @@
...
+                    initialized.store(true)
+                    sessionId?.let { onSessionInitialized?.invoke(it) }
+                }
+            } else {
+                if (!validateSession(call) || !validateProtocolVersion(call)) return
+            }
+
+            val hasRequest = messages.any { it is JSONRPCRequest }
+            if (!hasRequest) {
+                call.respondNullable(status = HttpStatusCode.Accepted, message = null)
```
dvilker, line 346: Perhaps because of the `Accept` request header, this line returns `406 Not Acceptable` instead of `202`. MCP Inspector throws an error in log, and the Python client crashes because of this. For me, `call.respondBytes(status = HttpStatusCode.Accepted, bytes = ByteArray(0))` helped.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -64,6 +66,51 @@ public fun Application.mcp(block: ServerSSESession.() -> Server) {
...
+public fun Application.mcpStatelessStreamableHttp(
+    enableDnsRebindingProtection: Boolean = false,
+    allowedHosts: List<String>? = null,
+    allowedOrigins: List<String>? = null,
+    eventStore: EventStore? = null,
+    block: RoutingContext.() -> Server,
+) {
+    routing {
+        post("/mcp") {
```
dvilker, line 118: `get` requests also need to be processed.

dvilker (commented):
I tried your PR in practice in Stateless mode and brought back what I found.

mantisMolloy:
Hi @zshea 

Thanks for this PR.

In relation to:
> But the response is always 200, even if we set the status explicitly. And the connection won't be closed with a 4xx or 5xx response. It breaks the MCP spec and can't be used directly. I think we need to create issues to upstream.

Was an issue created on ktor for this?

mantisMolloy:
One thing that's not clear to me

>  I can't write an integration test for this transport. The client StreamHttpTransport assumes the SSE stream while the server assumes the JSON response.

Looking at the spec the client MUST be able to accept a json response. Is this a bug in the client?

From the spec:

> If the input contains any number of JSON-RPC requests, the server MUST either return Content-Type: text/event-stream, to initiate an SSE stream, or Content-Type: application/json, to return one JSON object. The client MUST support both these cases.

mantisMolloy:
I played around with this a bit.

1. When using `enableJsonResponse = true` the client `StreamHttpTransport` works when you add a `sse` route for get requests using the ktor plugin. Just pass the session into the the server `StreamHttpTransport` and then the it will return a `406` status and this overcomes the issue where `the client assumes the SSE stream`. We should be able to add integration tests if the sse route is added.

2. I'm not sure if a fix is required upstream for the ktor routing plugin. Could we implement a version of `io.ktor.server.sse.ServerSSESession` and initiate the event stream inside `handlePostRequest` when `enableJsonResponse = false`.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
zshea: I didn't implement because it's not needed for json response. Once we add SSE back, we can do it then.

zshea:
> Hi @zshea
> 
> Thanks for this PR.
> 
> In relation to:
> 
> > But the response is always 200, even if we set the status explicitly. And the connection won't be closed with a 4xx or 5xx response. It breaks the MCP spec and can't be used directly. I think we need to create issues to upstream.
> 
> Was an issue created on ktor for this?

I found an answer here: https://youtrack.jetbrains.com/issue/KTOR-8327/SSE-Document-how-to-receive-SSE-requests-with-POST-and-other-methods#focus=Comments-27-12594500.0-0

The workaround significantly changes the architectureof  how we implement transport: need to split header response and body response.

zshea:
> Thank you @zshea for the PR Would it be possible to add integration tests to make sure it actually works? 

TBH I've been struggling with adding tests here, would appreciate some help here, or I can dive more later. Sorry almost forgot this PR for a while...

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
dvilker: > I didn't implement because it's not needed for json response. Once we add SSE back, we can do it then.

https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http

_The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint_

Without processing (which responds with code 405), it seems the inspector was spamming errors.

I replaced `post(...` with `route("...`. Everything else has already been implemented by you.

zacutti:
> > Thank you @zshea for the PR Would it be possible to add integration tests to make sure it actually works?
> 
> TBH I've been struggling with adding tests here, would appreciate some help here, or I can dive more later. Sorry almost forgot this PR for a while...

Any ETA for this feature? how can I help?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
joewlambeth: I'm drafting a PR to return 405 for the stateless extension on GET requests, as per spec

> The server MUST either return Content-Type: text/event-stream in response to this HTTP GET, or else return HTTP 405 Method Not Allowed, indicating that the server does not offer an SSE stream at this endpoint.

joewlambeth:
Thanks @devcrocod for the work on resolving the merge conflicts. I'm creating integration tests for this, but I'm running into an issue where Ktor is sending 406 Not Acceptable on the Streamable Server Transport for the initialize call, despite an OK payload. There shouldn't be any issues serializing the JSONRPCResponse. This wasn't happening with OldSchema integration tests, but now it seems to be happening. I'll look into this more soon, but I'll leave some details here in case there's some obvious I'm missing.

The attempted JSON RPC payload is as follows:
```
JSONRPCResponse(id=NumberId(value=1), result=InitializeResult(protocolVersion=2025-06-18, capabilities=ServerCapabilities(tools=Tools(listChanged=true), resources=null, prompts=null, logging=null, completions=null, experimental=null), serverInfo=Implementation(name=test-server, version=1.0, title=null, websiteUrl=null, icons=null), instructions=null, meta=null))
```

However, this is what the Ktor server is sending back. 
```
[DefaultDispatcher-worker-2 @request-handler#23] INFO io.ktor.server.Application - Status: 406 Not Acceptable, Method: POST, Path: /mcp
```

Here's how i'm starting the test server
```kotlin
            TransportKind.STREAMABLE_HTTP_STATELESS -> {
                serverEngine = embeddedServer(ServerCIO, host = host, port = port) {
                    install(CallLogging) {
                        level = Level.INFO 
                        format { call ->
                            val status = call.response.status()
                            val httpMethod = call.request.httpMethod.value
                            val path = call.request.path()
                            "Status: $status, Method: $httpMethod, Path: $path"
                        }
                    }
                    routing {
                        mcpStatelessStreamableHttp { server }
                    }
                }.start(wait = false)
            }
```

devcrocod:
Hey @joewlambeth,

I didn’t open a separate pull request and instead continued the work in the same branch
I felt it would be easier to keep track of the progress that way. I hope this didn’t cause any inconvenience for anyone.

While going through the review of this PR, I fixed a number of issues I discovered both in the streamable implementation itself and in our protocol. Next, I’m planning to polish the implementation, review the KtorServer logic, and write an integration test using the typescript client against our server to verify that the basic streamable transport flow works correctly. I think this should be enough to merge these changes.

After that, we’ll add broader test coverage for the streamable server transport, as well as [conformance tests](https://github.com/modelcontextprotocol/conformance), to make sure we fully align with the specification.

Thanks a lot for the example you provided, it’s definitely helpful. We also downgraded the Ktor version recently, which could be another reason why your test stopped working. I’ll take a closer look at this when I work on the integration test with the typescipt client

joewlambeth:
A few issues I discovered that finally fixed my integration test issues. 

1. I had to include the `ContentNegotiation` plugin for the server. I need to see what's the best way to communicate this requirement in the docs, if required. This also requires a couple of extra KTOR dependencies to be pulled in the server artifact.
2. I need to add a GET route for `/mcp` that automatically returns 405 Method Not Allowed. [This is per spec](https://modelcontextprotocol.io/specification/draft/basic/transports#listening-for-messages-from-the-server). 
3. With how the payload was being serialized, Kotlinx Serialization was failing. I've redone how the serialization is done on the response.

[I've raised a PR](https://github.com/zshea/kotlin-sdk/pull/1) to @zshea's fork. I'm hoping to make some additional changes, I can possible work on the conformance tests in the future!

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/jsonRpc.kt
```diff
@@ -85,6 +85,12 @@ public sealed interface JSONRPCMessage {
     public val jsonrpc: String
 }
 
+@Serializable
+public data object PrimingEventMessage : JSONRPCMessage {
+    @EncodeDefault
+    override val jsonrpc: String = JSONRPC_VERSION
+}
+
```
kpavlov, line 93: what is it for?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -0,0 +1,674 @@
...
+    private val enableDnsRebindingProtection: Boolean = false,
+    private val allowedHosts: List<String>? = null,
+    private val allowedOrigins: List<String>? = null,
+    private val eventStore: EventStore? = null,
+    private val retryIntervalMillis: Long? = null,
+) : AbstractTransport() {
+    public var sessionId: String? = null
+        private set
+
+    private var sessionIdGenerator: (() -> String)? = { Uuid.random().toString() }
```
kpavlov, line 124:
```suggestion
    private var sessionIdGenerator: (() -> String)? = { Uuid.random().toHexString() }
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -0,0 +1,674 @@
...
+
+        val isTerminated = message is JSONRPCResponse || message is JSONRPCError
+        if (!isTerminated) return
+
+        requestToResponseMapping[responseRequestId!!] = message
+        val relatedIds = requestToStreamMapping.filterValues { it == streamId }.keys
+
+        if (relatedIds.any { it !in requestToResponseMapping }) return
+
+        streamMutex.withLock {
```
kpavlov, line 218: It's dangerous to keep the lock while doing IO operation: IO takes a lot of time. I see  `activeStream.call.respond(payload)` is executed under the lock.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -0,0 +1,674 @@
...
+                        "Invalid Request: Only one initialization request is allowed",
+                    )
+                    return
+                }
+
+                sessionMutex.withLock {
+                    if (sessionId != null) return@withLock
+                    sessionId = sessionIdGenerator?.invoke()
+                    initialized.store(true)
+                    sessionId?.let { onSessionInitialized?.invoke(it) }
```
kpavlov, line 338: shuould it be under the lock?

kpavlov:
Changes made in #449

PR #243, opened 2025-08-28
Add recovery during message deserialization and update workflow to up…

skarpovdev:
…load tests result

## Motivation and Context

Now `should ignore first output messages` test fails because the valid JSON message is on the same line as non-JSON "Stdio server started".
I added a recovery which attempts to find JSON by searching for the first `{`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
@@ -44,7 +44,17 @@ public class ReadBuffer {
         try {
             return deserializeMessage(line)
         } catch (e: Exception) {
-            logger.error(e) { "Failed to deserialize message from line: $line\nSkipping..." }
+            logger.error(e) { "Failed to deserialize message from line: $line\nAttempting to recover..." }
+            // if there is a non-JSON object prefix, try to parse from the first '{' onward.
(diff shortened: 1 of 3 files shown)


kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
```diff
@@ -44,7 +44,17 @@ public class ReadBuffer {
         try {
             return deserializeMessage(line)
         } catch (e: Exception) {
-            logger.error(e) { "Failed to deserialize message from line: $line\nSkipping..." }
+            logger.error(e) { "Failed to deserialize message from line: $line\nAttempting to recover..." }
+            // if there is a non-JSON object prefix, try to parse from the first '{' onward.
+            val braceIndex = line.indexOf('{')
```
tiginamaria, line 49: Is it because you can have a text + json in the output? If so, great fix!

tiginamaria (approved):
Thank you! I've left just one clarification question there to know the exact reason of the parsing error. Other than that, all is good!

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
skarpovdev: Correct, in case of the aforementioned test, the resulting string is this:
image

Issue #158, opened 2025-07-18
Support resource links in tool result content


ptitjes:
**Is your feature request related to a problem? Please describe.**
Protocol 2025-06-18 allows a new type `resource_link` of tool result content: https://modelcontextprotocol.io/specification/2025-06-18/server/tools#resource-links

It is only provided for tool results and not for prompt results, so supporting this might need to modify the `PromptMessageContent` sealed hierarchy.

**Describe the solution you'd like**
- Make a new `ToolResultContent` sealed hierarchy that share sub-classes with the `PromptMessageContent` sealed hierarchy.
- Add a new `ResourceLinkContent` subclass to that new sealed hiearchy.

**Describe alternatives you've considered**
N/A

**Additional context**
- [Protocol 2025-06-18 > Key Changes](https://modelcontextprotocol.io/specification/2025-06-18/changelog)
- [Tools > Data Types > Tool Result > Resource Links](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#resource-links)

ptitjes:
I can PR for this.

Issue #185, opened 2025-07-21
Revisit schema. Generate Kotlin code from json-schema


devcrocod:
Currently, the model classes are written by hand and not in sync with the specification.

The good news is that there is a [Json-Schema](https://github.com/modelcontextprotocol/modelcontextprotocol/tree/main/schema) for the protocol and pojos can be generated automatically.

# ToDo
- Consider [openapi-generator gradle plugin](https://github.com/OpenAPITools/openapi-generator) to generate model classes. Generate classes to `io.modelcontextprotocol.kotlin.sdk.model`
- Remove generated classes out of [types.kt](https://github.com/modelcontextprotocol/kotlin-sdk/blob/main/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt). `types.kt` has degraded; it has become too huge. Consider extending hand-written classes from generated ones, marking hand-written classes as deprecated and remove them completely on the next releae
- Move serializers to a separate package, e.g. `io.modelcontextprotocol.kotlin.sdk.model.serializers`.
- Add comprehensive JSON serialization/deserialization tests

# DoD
- Model classes are generated as a build step from the JSON schema
- Unit tests for JSON serialization exist and are green. Test coverage is 100% for model classes. Take sample jsons from the latest [specification](https://modelcontextprotocol.io/specification/2025-06-18)

PR #244, opened 2025-09-04
158 support resource links

dstibbe:
Support resource links in tool result content #158

* Refactored  PromptMessageContent to ContentBlock
* Added the ResourceLink as a ContentBlock type

## Motivation and Context
The new spec introduced a new `ContentBlock` type: `ResourceLink`. This type however hasn't been implemented yet.
Noticed that there was actually no `ContentBlock` type at all in the core as the latest schema reference specified. Instead there is a `PromptMessageContent`. 

## How Has This Been Tested?
Added unit tests. Ran all tests.
Also used the build in a standalone project:
* calling a tool returning a ResourceLink via de client directly
* called a tool using Jetbrains Koog framework

## Breaking Changes
Added the following:
* `ContentBlock`
* `CreateMessageResultContent`, and
*  `SamplingMessageContent`

However, I kept the two intrefaces `PromptMessageContent` and `PromptMessageContentMultimodal` purely for backwards compatibility. They are deprecated however.

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)
- [ x ] Documentation update

## Checklist
- [ x ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ x ] My code follows the repository's style guidelines
- [ x ] New and existing tests pass locally
- [ x ] I have added appropriate error handling  - none needed
- [ x ] I have added or updated documentation as needed

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -40,7 +40,7 @@ jobs:
 
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@v3
+        uses: github/codeql-action/init@v4
         with:
           languages: ${{ matrix.language }}
           build-mode: manual
@@ -64,6 +64,6 @@ jobs:
             --no-daemon --stacktrace --rerun-tasks
 
       - name: Analyze
-        uses: github/codeql-action/analyze@v3
+        uses: github/codeql-action/analyze@v4
         with:
           category: '/language:${{ matrix.language }}'
--- a/kotlin-sdk-core/build.gradle.kts
+++ b/kotlin-sdk-core/build.gradle.kts
@@ -7,7 +7,7 @@ plugins {
     id("mcp.publishing")
     id("mcp.dokka")
     alias(libs.plugins.kotlinx.binary.compatibility.validator)
-    id("org.openapi.generator") version "7.15.0"
+    id("org.openapi.generator") version "7.16.0"
 }
 
 /*


kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/TypesTest.kt
```diff
@@ -149,15 +149,61 @@ class TypesTest {
...
+            size = 76859L,
+            title = "This is a pdf",
+            uri = "file:///path/to/my_file.pdf",
+        )
+
+        val json = McpJson.encodeToString<ContentBlock>(resourceLink)
+        val decoded = McpJson.decodeFromString<ContentBlock>(json)
+
+        assertIs<ResourceLink>(decoded)
+        with(resourceLink) {
```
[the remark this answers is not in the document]
dstibbe: Fixed

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -970,19 +970,34 @@ public data class GetPromptRequest(
...
+@Deprecated("For backwards compatibility; use ContentBlock instead", ReplaceWith("ContentBlock"))
 public sealed interface PromptMessageContent {
     public val type: String
 }
 
+@Deprecated(
+    "For backwards compatibility; use CreateMessageResultContent or SamplingMessageContent instead",
+    ReplaceWith("CreateMessageResultContent"),
+)
+public sealed interface PromptMessageContentMultimodal : PromptMessageContent
```
kpavlov, line 997: Would it be possible to add KDoc to all public methods?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -970,19 +970,34 @@ public data class GetPromptRequest(
...
-@Serializable(with = PromptMessageContentMultimodalPolymorphicSerializer::class)
-public sealed interface PromptMessageContentMultimodal : PromptMessageContent
+@Serializable(with = ContentBlockPolymorphicSerializer::class)
+public sealed interface ContentBlock : PromptMessageContent
+
+/**
+ * Represents content for the CreateMessageResult
+ */
+@Serializable(with = CreateMessageResultContentMultimodalPolymorphicSerializer::class)
+public sealed interface CreateMessageResultContent : ContentBlock
```
kpavlov: CreateMessageResult supports only `TextContent | ImageContent | AudioContent`. ContentBlock has also `ResourceLink
| EmbeddedResource`

https://modelcontextprotocol.io/specification/2025-06-18/schema#createmessageresult

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -970,19 +970,34 @@ public data class GetPromptRequest(
...
+
+/**
+ * Represents content for the SamplingMessage
+ */
+@Serializable(with = SamplingMessageContentMultimodalPolymorphicSerializer::class)
+public sealed interface SamplingMessageContent : ContentBlock
```
kpavlov: SamplingMessage supports only `TextContent | ImageContent | AudioContent`

https://modelcontextprotocol.io/specification/2025-06-18/schema#samplingmessage

kpavlov (commented):
#294 introduces model generation from MCP json schema. It should solve the issue of models are out of sync with the schema.

I would push #294 forward and then migrate to generated models, so this PR would not be necessary.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -975,19 +975,45 @@ public data class GetPromptRequest(
...
+    "For backwards compatibility; use CreateMessageResultContent or SamplingMessageContent instead",
+    ReplaceWith("CreateMessageResultContent"),
+)
 public sealed interface PromptMessageContentMultimodal : PromptMessageContent
 
+/**
+ * Represents the types of a ContentBlock
+ */
+@Serializable(with = ContentBlockPolymorphicSerializer::class)
+public sealed interface ContentBlock : PromptMessageContent
```
devcrocod, line 1003: You’re introducing a new `ContentBlock` interface, but still extending a deprecated one.
In my opinion, that’s not the best approach

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -975,19 +975,45 @@ public data class GetPromptRequest(
...
+ * Represents content for the CreateMessageResult
+ */
+@Serializable(with = CreateMessageResultContentMultimodalPolymorphicSerializer::class)
+public sealed interface CreateMessageResultContent : ContentWithType
+
+/**
+ * Represents content for the SamplingMessage
+ */
+@Serializable(with = SamplingMessageContentMultimodalPolymorphicSerializer::class)
+public sealed interface SamplingMessageContent : ContentWithType
```
devcrocod, line 1015: It’s not clear to me why these interfaces were separated and added, what’s the purpose?
Also, since custom serializers were defined for them, wouldn’t deserializers be required as well?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -1002,7 +1028,9 @@ public data class TextContent(
      * Optional annotations for the client.
      */
     val annotations: Annotations? = null,
-) : PromptMessageContentMultimodal {
+) : ContentBlock,
```
devcrocod, line 1031: Wouldn’t it be possible to just inherit only from `ContentBlock`?

devcrocod (changes requested):
I have a few questions regarding this PR.
Could you please take a moment to answer them?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
dstibbe: The reason was to keep it backwards compatible. To avoid it immediately breaking ppl's builds when they are upgrading

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
dstibbe: Because, although similar, schema wise SamplingMessage.content is not the CreateMessageResult.content. 
Separating them prevents having the issue that occurred now (not being able to simply add resourcelink to contentblock) because things were treated the same while they were not.

The serializers contain the deserializer selection ?

Regardless, this discussion is moot  :P  This PR will be removed and #294 is the future ...

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
dstibbe: Not if someone was actually using PromptMessagecontentMultimodal. Thus kept that for backwards compatibility. Till the deprecated classes would be removed

dstibbe: Thanks for taking the trouble of reviewing my code btw. Appreciated.

dstibbe:
> I have a few questions regarding this PR. Could you please take a moment to answer them?

So, to be clear,  the approach I took was to add the content definitions of the 'new' (2025-06-18) schema while not immediately breaking it for anyone upgrading. Such that I could get the resourcelink in the tool call result as quick as possible and get some work done with it. While the SDK project could then softly phase out the deprecated classes.

Alas, #294 is the way of the future, which means I will probably will have to wait a while till an official release with resourcelink will be available...  
I will be using a custom build till then I guess.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
devcrocod: In the future, according to the deprecation cycle, we’ll eventually have to remove the old interfaces, which will lead to incompatibility issues.
I see two possible approaches here:
- To maintain compatibility, we can create new classes and mark the old ones as deprecated, ensuring they don’t depend on each other.
- Alternatively, we can completely rewrite the old ones, which will cause incompatible, but not binary incompatibility. I believe this option is better since any issues will be detected at compile time

devcrocod:
> Alas, https://github.com/modelcontextprotocol/kotlin-sdk/pull/294 is the way of the future, which means I will probably will have to wait a while till an official release with resourcelink will be available...

Unfortunately, the openapi-generator doesn’t fully meet our needs, see https://github.com/modelcontextprotocol/kotlin-sdk/pull/294#pullrequestreview-3278839943

Right now, we’re only using it to compare the diff between our schema implementation and the specification, to identify what we might have missed.

At the same time, I still have #185 in my plans. I want our implementation to fully comply with the schema and to improve serialization, since the current custom serialization is redundant

dstibbe:
> > Alas, #294 is the way of the future, which means I will probably will have to wait a while till an official release with resourcelink will be available...
> 
> Unfortunately, the openapi-generator doesn’t fully meet our needs, see [#294 (review)](https://github.com/modelcontextprotocol/kotlin-sdk/pull/294#pullrequestreview-3278839943)
> 
> Right now, we’re only using it to compare the diff between our schema implementation and the specification, to identify what we might have missed.
> 
> At the same time, I still have #185 in my plans. I want our implementation to fully comply with the schema and to improve serialization, since the current custom serialization is redundant
 
Ok . If that is the case, this means this PR CAN go through ? 
Especially in view of #378 .

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
dstibbe: Of course you will have to remove old interfaces, but this way it allows the new interfaces to be used as soon as possible, while breaking nothing. And eventually remove the deprecated interfaces when you increment the major version.
I believe "As soon as possible" should be kept in mind, since this SDK is lagging more and more behind to the official schema.
If it is possible without too much effort I'd say that would be the preferred approach. Now, if the new schema introduces something that is completely incompatible with the old schema, sure, break it.

devcrocod:
We have now added the [resource links](https://github.com/modelcontextprotocol/kotlin-sdk/blob/13f858d5d24a1e8a07d06a200bab32a4a640cc6f/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/content.kt#L127) and fully implemented the latest version of the specification schema, including restructuring the types.

Given this, the required changes are already covered, so I’m going to close this PR. 
Thanks for your contribution!

PR #267, opened 2025-09-15
Mark `testMultipleClientParallel` as ignored due to flakiness

devcrocod:
Ignore flaky `testMultipleClientParallel`


kpavlov (approved):
It's not good. Let's address the root cause separately

kpavlov (approved):

PR #271, opened 2025-09-15
Publich test report

kpavlov:
Update build workflow: refine report handling and add test result publishing

- Change `Upload Reports` to run only when not cancelled.
- Add a new step to publish test results with a detailed summary.

## Motivation and Context

The test report should be easy to find

## Types of changes

- [ ] Documentation update

## Checklist


devcrocod (approved):
lgtm

.github/workflows/build.yml
```diff
@@ -37,13 +37,24 @@ jobs:
           JAVA_OPTS: "-Xmx8g -Dfile.encoding=UTF-8 -Djava.awt.headless=true -Dkotlin.daemon.jvm.options=-Xmx6g"
 
       - name: Upload Reports
-        if: always()
```
skarpovdev, line 40: Argument for keeping it - there was a case when our tests that use JUnit5 weren't running because of the misconfiguraion, so we didn't know which tests are run and which are not.
Not a strong opinion so if there are any downsides in having it, let's switch to `${{ !cancelled() }}` indeed.

PR #278, opened 2025-09-17
Prerelease 0.7.2-alpha1

devcrocod:
## Motivation and Context
- test new publish plugin
- add package.kt to kotlin-sdk umbrella module to resolve klib generation errors during Maven publishing

## Breaking Changes
no

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,7 +6,7 @@ plugins {
 
 allprojects {
     group = "io.modelcontextprotocol"
-    version = "0.7.2-SNAPSHOT"
+    version = "0.7.2-alpha1"
 }
 
 dependencies {
--- a/buildSrc/src/main/kotlin/mcp.publishing.gradle.kts
+++ b/buildSrc/src/main/kotlin/mcp.publishing.gradle.kts
@@ -5,7 +5,6 @@ plugins {
 }
 
 mavenPublishing {
-    signAllPublications()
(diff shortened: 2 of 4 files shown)


kpavlov (dismissed):
👍

kpavlov (changes requested):
Thank you, @devcrocod, for tackling the publication work — I really appreciate the effort.

I’d like to ask whether we can remove the kotlin-sdk module now instead of keeping it. I understand why it was introduced and the maintenance involved, but in practice, it tends to:

- It can unintentionally pull server dependencies into client applications (and vice versa), which is confusing for users.
- It commonly ends up as an unused transitive dependency.
- It increases Gradle configuration complexity without delivering clear value after the client/server split.

Could we make a separate PR that focuses on removing the module and the related Gradle/config changes? We could also keep the version bump in its own PR so each change has a single, clear purpose.

devcrocod:
@kpavlov 
I’d like to keep the kotlin-sdk until there’s a clear need to remove it

- we have a choice: the user can use either kotlin-sdk, or client, or server
- we maintain compatibility with older releases where there was only one kotlin-sdk
- this simplifies sdk usage for new users
- we stay consistent with other sdks that provide a single artifact

kpavlov (approved):
Let's merge to unblock the release, but I am not fully agree with your arguments

PR #288, opened 2025-09-24
Implement logging disable feature in MCP server

JGuck:
Added a method to disable logging in the MCP server.

Add a function to disable the logger. 

## Motivation and Context
This is needed because prints are causing errors on my nodejs stdio implementation

## How Has This Been Tested?
Compiles and works in my environment

## Breaking Changes
no

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist


kpavlov:
Thank you, @JGuck 
Wouldn't it be more appropriate to configure logging levels via logging provider, e.g., SLF4J, instead of pulling it into the library?
I am curious about your use case and where the current logging pattern doesn't work for you.

JGuck:
@kpavlov Yes you're right. In my first investigation, I realized that there is no provider for js . But KotlinLoggingConfiguration.logLevel = Level.OFF is doing the trick for js.

This only works if done in mainJs. On common code it is not existing. That's w I was not able to find it in the first place.

PR #290, opened 2025-09-25
Adds 'instructions' string to the Server and propagates to the InitializeResult

jclyne:
## Motivation and Context
The [schema](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/cca15204f62143bc48429ece62b6e35b06d9785b/schema/2025-03-26/schema.ts#L192) for the InitializeResult contains an optional 'instructions' string

## How Has This Been Tested?
Its relatively simple change with unit tests included

## Breaking Changes
No, field is optional and defaults to null

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context
The instructions string is included in the schema for for version "2025-03-26"

The instructions in the Server class are given as a provider function with the result of the provider propagated to the underlying ServerSession. As the ServerSession change is meant to allow for a 'Singleton scoped' server class, this provides the ability to dynamically modify the instructions at the server level but have them be immutable per session.
 
This also refactors the `serverInfo` and `instructions` properties to be protected. This `Server` class is open and meant to be overridden, there is no harm in having these immutable properties available to subclasses. 

This adds the serverInstructions to the client as returned from the InitializeResult. This is used to verify the response in the Server tests.


kotlin-sdk-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerTest.kt
```diff
@@ -471,4 +471,58 @@ class ServerTest {
         }
         assertEquals("Server does not support resources capability.", exception.message)
     }
+
+    @Test
+    fun `Server constructor should accept instructions parameter`() = runTest {
```
kpavlov, line 496: Thank you for the test!
One question: Is it possible to verify in the test that the server has received the instructions string?
Also, the test could be parametrized

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/TypesTest.kt
```diff
@@ -394,4 +394,63 @@ class TypesTest {
...
+    fun `should serialize and deserialize InitializeResult with instructions`() {
+        val serverInfo = Implementation(name = "test-server", version = "1.0.0")
+        val instructions = "This server provides file system access. Use the 'read' tool to read files."
+        val result = InitializeResult(
+            serverInfo = serverInfo,
+            instructions = instructions,
+        )
+
+        val json = McpJson.encodeToString(result)
+        val decoded = McpJson.decodeFromString<InitializeResult>(json)
```
kpavlov, line 435: nit: I think we should introduce some serialization helper for tests like this, like [here](https://github.com/JetBrains/koog/blob/develop/test-utils/src/commonMain/kotlin/ai/koog/test/utils/SerializationHelpers.kt)

jclyne: This would probably be applicable to all the tests in this file, so I would prefer to leave it to a different PR if that's ok

kotlin-sdk-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerTest.kt
jclyne: The serverInstructions were added to the client as a property that can be verified in these tests.

kpavlov (dismissed):
LGTM

kyeotic:
@kpavlov We also need this for parity with FastMCP python servers. Any chance you can ballpark an ETA for this getting merged/released?

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/TypesTest.kt
devcrocod: I agree with both statements

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -53,8 +53,15 @@ public class ServerOptions(public val capabilities: ServerCapabilities, enforceS
...
  * @param options Configuration options for the server.
+ * @param instructionsProvider Optional provider for instructions from the server to the client about how to use
+ * this server. The provider is called each time a new session is started to support dynamic instructions.
  */
-public open class Server(private val serverInfo: Implementation, private val options: ServerOptions) {
+
+public open class Server(
+    protected val serverInfo: Implementation,
+    protected val options: ServerOptions,
+    protected val instructionsProvider: (() -> String)? = null,
```
devcrocod, line 63: Why is this implemented as a lambda and not just a string?

devcrocod (changes requested):
Thank you for your contribution! 

I have a question regarding the api design, could you please take a look? 
Everything else looks good to me!

devcrocod:
Hi @jclyne 
We recently added ServerSession, which still requires some improvements. 
Soon we plan to estimate the timeline for a stable release.

In the meantime, next week we’d like to set up snapshot releases

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
jclyne: This is allowing the instructions to potentially be dynamically loaded from a database or string server. With the change to the ServerSession, the Server seems to take more of a Singleton scope vs a Server per stream. The idea is that the Server has a provider method for the string which it evaluates as a string value for the ServerSession.  

~Its possible to add an additional constructor that just takes a static string and converts it to a lamda, but that is trivial for the caller.~

I went ahead and added an alternative constructor that takes a static string

jclyne:
> Hi @jclyne We recently added ServerSession, which still requires some improvements. Soon we plan to estimate the timeline for a stable release.
> 
> In the meantime, next week we’d like to set up snapshot releases

Thanks @devcrocod!  I would like to get this in the snapshot release if at all possible.

devcrocod (approved):
LGTM!

I would prefer the primary constructor to be as simple as possible, but now it's okay

Please update api, then we'll be able to merge these changes

devcrocod (approved):

Issue #302, opened 2025-10-04
Make sample projects compilable standalone


kpavlov:
**Describe the bug**
Currently sample projects are viable only as a submodules of the main project. 

**To Reproduce**
1. User clones the repo
2. cd samples/(project)
3. Build gradle project- it fails

**Expected behavior**
1. git clone
2. cd kotlin-sdk/samples/PROJECT
3. ./gradlew build
4. Success 🎉 

**Logs**
If applicable, add logs to help explain your problem.

```
./gradlew kotlin-mcp-server:runJvm                                                                                                                                                 

Welcome to Gradle 8.14.3!
Starting a Gradle Daemon (subsequent builds will be faster)

> Task :kotlin-mcp-server:compileKotlinJvm FAILED
e: file:///Users/user/TestWorkspace/kotlin-sdk/samples/kotlin-mcp-server/src/commonMain/kotlin/server.kt:26:50 Unresolved reference 'ServerSession'.
e: file:///Users/user/TestWorkspace/kotlin-sdk/samples/kotlin-mcp-server/src/commonMain/kotlin/server.kt:98:48 Unresolved reference 'ServerSession'.
e: file:///Users/user/TestWorkspace/kotlin-sdk/samples/kotlin-mcp-server/src/commonMain/kotlin/server.kt:116:31 Unresolved reference 'onClose'.
e: file:///Users/user/TestWorkspace/kotlin-sdk/samples/kotlin-mcp-server/src/commonMain/kotlin/server.kt:124:60 Unresolved reference 'transport'.
```

**Additional context**

PR #330, opened 2025-10-22
#302 Update samples build workflow

kpavlov:
# #302 Update samples build workflow

- Introduce workflow to build samples independently from the main project
- Keep building samples against the current snapshot version of the SDK. Update composite build configuration.
- Temporarily disable "samples/kotlin-mcp-server": the multiplatform project is broken. **It doesn't compile for me even on the main branch**. Not working with [MCP Instepcor](https://modelcontextprotocol.io/docs/tools/inspector), because CORS is not configured. Deserves a separate PR to fix.

## Motivation and Context

Samples should be independently buildable

## Types of changes

- [ ] Documentation update
- [x] Build improvement
- [x] Tests

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,9 +49,9 @@ jobs:
         working-directory: ./samples/kotlin-mcp-client
         run: ./gradlew --no-daemon clean build -Pmcp.kotlin.overrideVersion=0.0.1-SNAPSHOT
 
-      - name: Build Kotlin-MCP-Server Sample
-        working-directory: ./samples/kotlin-mcp-server
-        run: ./gradlew --no-daemon clean build -Pmcp.kotlin.overrideVersion=0.0.1-SNAPSHOT
+#      - name: Build Kotlin-MCP-Server Sample
+#        working-directory: ./samples/kotlin-mcp-server
+#        run: ./gradlew --no-daemon clean build -Pmcp.kotlin.overrideVersion=0.0.1-SNAPSHOT
 
       - name: Build Weather-Stdio-Server Sample
         working-directory: ./samples/weather-stdio-server
--- a/.github/workflows/samples.yml
+++ b/.github/workflows/samples.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         sample:
           - kotlin-mcp-client
-          - kotlin-mcp-server
+#          - kotlin-mcp-server
           - weather-stdio-server
 
     name: Build Sample
--- a/samples/kotlin-mcp-client/settings.gradle.kts
+++ b/samples/kotlin-mcp-client/settings.gradle.kts
@@ -17,7 +17,7 @@ dependencyResolutionManagement {
                 "mcp.kotlin.overrideVersion",
             ).orNull
             if (mcpKotlinVersion != null) {
-                logger.lifecycle("Using the override version $mcpKotlinVersion of KotlinMCP SDK")
+                logger.lifecycle("Using the override version $mcpKotlinVersion of MCP Kotlin SDK")
                 version("mcp-kotlin", mcpKotlinVersion)
             }
         }
--- a/samples/kotlin-mcp-server/settings.gradle.kts
+++ b/samples/kotlin-mcp-server/settings.gradle.kts
@@ -17,7 +17,7 @@ dependencyResolutionManagement {
                 "mcp.kotlin.overrideVersion",
             ).orNull
             if (mcpKotlinVersion != null) {
-                logger.lifecycle("Using the override version $mcpKotlinVersion of KotlinMCP SDK")
+                logger.lifecycle("Using the override version $mcpKotlinVersion of MCP Kotlin SDK")
                 version("mcp-kotlin", mcpKotlinVersion)
             }
         }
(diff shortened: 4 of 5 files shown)


samples/kotlin-mcp-client/settings.gradle.kts
```diff
@@ -1,20 +1,25 @@
 rootProject.name = "kotlin-mcp-client"
 
-pluginManagement {
-    repositories {
-        mavenCentral()
-        gradlePluginPortal()
-    }
+plugins {
+    // Apply the foojay-resolver plugin to allow automatic download of JDKs
+    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
EugeneTheDev, line 5: Consider moving dependency declaration to the `libs.versions.toml`

samples/kotlin-mcp-server/settings.gradle.kts
```diff
@@ -1,20 +1,25 @@
 rootProject.name = "kotlin-mcp-server"
 
-pluginManagement {
-    repositories {
-        mavenCentral()
-        gradlePluginPortal()
-    }
+plugins {
+    // Apply the foojay-resolver plugin to allow automatic download of JDKs
+    id("org.gradle.toolchains.foojay-resolver-convention") version "1.0.0"
```
EugeneTheDev, line 5: Same, let's move to `libs.versions.toml`

samples/weather-stdio-server/src/test/kotlin/io/modelcontextprotocol/sample/client/FakeTest.kt
```diff
@@ -0,0 +1,11 @@
+package io.modelcontextprotocol.sample.client
+
+import kotlin.test.Test
+
+class FakeTest {
```
EugeneTheDev, line 5: Is it needed?

samples/kotlin-mcp-client/settings.gradle.kts
devcrocod: This is used in `settings.gradle.kts`, and I don’t think gradle can pull versions from the toml file at that stage though it’s worth checking to be sure

.github/workflows/build.yml
```diff
@@ -36,24 +36,26 @@ jobs:
...
-          ./gradlew --no-daemon --rerun-tasks \
+          ./gradlew --no-daemon \
+            --rerun-tasks \
             clean \
             ktlintCheck \
             build \
             koverLog koverHtmlReport \
-            publishToMavenLocal
+            publishToMavenLocal \
+            -Pversion=0.0.1-SNAPSHOT
```
devcrocod, line 46: Why is this version being used? Shouldn’t we use the latest version + snapshot instead?

samples/weather-stdio-server/src/test/kotlin/io/modelcontextprotocol/sample/client/FakeTest.kt
kpavlov: Junit fails if it discovers no tests. The proper test should be added separately

.github/workflows/samples.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       matrix:
         sample:
           - kotlin-mcp-client
-          - kotlin-mcp-server
+#          - kotlin-mcp-server
```
devcrocod, line 26: Why exclude the mcp-server example?

samples/kotlin-mcp-client/gradle/libs.versions.toml
devcrocod, line 1: Do we need a separate toml file for each example?
Maybe it would be clearer to specify the versions directly in the build.gradle files instead?

Also, we should check the dependabot workflow to make sure it monitors these files as well

settings.gradle.kts
```diff
@@ -26,8 +26,21 @@ include(
 )
 
 // Include sample projects as composite builds if this is the root project
-if (gradle.parent == null) {
-    includeBuild("samples/kotlin-mcp-client")
-    includeBuild("samples/kotlin-mcp-server")
-    includeBuild("samples/weather-stdio-server")
+// if (gradle.parent == null) {
+includeBuild("samples/kotlin-mcp-client") {
```
devcrocod, line 30: It’s not clear to me why we’re keeping the examples in composite builds, given that in this PR they’re being turned into separate projects

devcrocod (commented):
I have a few questions, could you please take a look at them?

samples/kotlin-mcp-client/gradle/libs.versions.toml
kpavlov: Yes, samples should be completely independent from SDK

settings.gradle.kts
kpavlov: IDEA project should contain sdk and samples. It is very valuable to have both sdk and samples in the same project. Composite builds provide this flexibility

.github/workflows/samples.yml
kpavlov: Because it's broken. See PR description

.github/workflows/build.yml
kpavlov: No, it would be harder to maintain. This is an ephemeral version for the build process only.

samples/kotlin-mcp-client/settings.gradle.kts
kpavlov: No, unfortunately, the version catalog is not available at this point :(

devcrocod (approved):
lgtm

PR #289, opened 2025-09-24
Add metadata support to callTool method

maeryo:
According to https://modelcontextprotocol.io/specification/2025-06-18/basic/index#meta, the _meta parameter should be passable when calling tools. While the CallToolRequest class (in types.kt) currently includes _meta in its definition, SDK users cannot actually input _meta when invoking the callTool method.

This Pull Request adds a metadata parameter to the CallTool method and includes functionality to convert various data formats without loss.

Additionally, to encourage the use of valid Key name formats as defined by the protocol, validation for prefixes has been implemented.

PR #333, opened 2025-10-23
Refactor MockTransport

kpavlov:
# Refactor MockTransport

This replaces bespoke test transport with a reusable, configurable MockTransport in a shared testing package.

The new MockTransport supports registering handlers for specific JSON-RPC methods (success and error), records sent/received messages, and provides an awaitMessage helper with polling and timeouts. Tests were added to validate handler registration, auto-responses, error responses, concurrency, and message awaiting behavior. Existing client tests were updated to configure MockTransport via lambdas instead of hardcoded logic. Additionally, coroutine test utilities were added to the core test dependencies.

## Motivation and Context

To provide better reusable test utilities

## How Has This Been Tested?
- [x] Unit tests

## Breaking Changes
No

## Types of changes

- [ ] Documentation update
- [x] Test improvement

## Checklist

## Additional context
#289 follow-up cc: @maeryo


kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+        mutex.withLock {
+            _sentMessages += message
+        }
+
+        // Auto-respond to using preconfigured request handlers
+        when (message) {
+            is JSONRPCRequest -> {
+                val response = requestHandlers.firstOrNull {
+                    it.first.invoke(message)
+                }?.second?.invoke(message)
```
maeryo, line 71: While ConcurrentSet is used for requestHandlers, the iteration in send() method is not protected by the mutex, which could lead to race conditions when handlers are added/removed concurrently.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+     * @param poolInterval The interval at which the function polls the received messages. Default is 50 milliseconds.
+     * @param timeout The maximum time to wait for a matching message. Default is 3 seconds.
+     * @param timeoutMessage The error message to throw when the timeout is reached.
+     * Default is "No message received matching predicate".
+     * @param predicate A predicate function that returns true if the message matches the criteria.
+     * @return The first JSON-RPC message that matches the predicate.
+     */
+    @OptIn(ExperimentalTime::class)
+    public suspend fun awaitMessage(
+        poolInterval: Duration = 50.milliseconds,
```
maeryo, line 194: The parameter name poolInterval appears to be a typo. The correct terminology is "pollInterval" (as in "polling"), not "pool".

maeryo (commented):
The `_sentMessages` and `_receivedMessages` lists grow unbounded, which can cause memory issues in long-running or high-volume tests.

**Suggestion**

```kotlin
private val maxMessages: Int = 1000  // configurable
private val _sentMessages = ArrayDeque<JSONRPCMessage>()
private val _receivedMessages = ArrayDeque<JSONRPCMessage>()

override suspend fun send(message: JSONRPCMessage) {
    mutex.withLock {
        _sentMessages.add(message)
        if (_sentMessages.size > maxMessages) {
            _sentMessages.removeFirst()
        }
    }
    // ...
}
```

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransportTest.kt
```diff
@@ -0,0 +1,575 @@
...
+            JSONRPCRequest(
+                id = RequestId.StringId("pool-test"),
+                method = "initialize",
+                params = buildJsonObject { },
+            ),
+        )
+
+        // Should work with custom pool interval
+        val message = transport.awaitMessage(
+            poolInterval = 10.milliseconds,
```
maeryo, line 245: The parameter name poolInterval appears to be a typo and should be pollInterval (consistent with the polling pattern). This is the same issue found in MockTransport.kt

maeryo:
I noticed a few areas that could enhance test maintainability and reliability (MockTransportTest.kt) :

### 1. **Code Duplication** (Lines 338-339, 374, 432, 472, 514, 549)
The same transport setup pattern is repeated across multiple tests. Consider extracting a helper:

```kotlin
private fun createCustomTransport(configure: MockTransport.() -> Unit = {}): MockTransport {
    return MockTransport(configure).apply { onMessage { } }
}
```

### 2. **Test Isolation** (Lines 29-48)
All tests share the same `transport` instance. Since `MockTransport` accumulates messages internally, this could cause test interference. Consider either:
- Adding a `clearMessages()` method and calling it in `@BeforeTest`, or
- Creating fresh transport instances per test when isolation is needed

```kotlin
// Remove shared transport property and create in each test
@Test
fun `some test`() = runTest {
    val transport = MockTransport {
        onMessageReplyResult(Method.Defined.Initialize) { ... }
    }
    transport.onMessage { }
    // test logic
}
```

### 3. **Redundant Assertions** (Throughout)
The pattern `assertNotNull(x)` + `assertTrue(x is Type)` is redundant in Kotlin. The type check already ensures non-null:

```kotlin
// Current
assertNotNull(message)
assertTrue(message is JSONRPCResponse)

// Suggested
assertTrue(message is JSONRPCResponse)  // Smart cast handles both checks
```

These are minor improvements but would help keep the test suite cleaner and more maintainable. WDYT?

kotlin-sdk-core/api/kotlin-sdk-core.api
```diff
@@ -3364,3 +3364,24 @@ public final class io/modelcontextprotocol/kotlin/sdk/shared/WebSocketMcpTranspo
...
+	public fun onError (Lkotlin/jvm/functions/Function1;)V
+	public fun onMessage (Lkotlin/jvm/functions/Function2;)V
+	public final fun onMessageReply (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
+	public final fun onMessageReplyError (Lio/modelcontextprotocol/kotlin/sdk/Method;Lkotlin/jvm/functions/Function1;)V
+	public static synthetic fun onMessageReplyError$default (Lio/modelcontextprotocol/kotlin/sdk/testing/MockTransport;Lio/modelcontextprotocol/kotlin/sdk/Method;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)V
+	public final fun onMessageReplyResult (Lio/modelcontextprotocol/kotlin/sdk/Method;Lkotlin/jvm/functions/Function1;)V
+	public fun send (Lio/modelcontextprotocol/kotlin/sdk/JSONRPCMessage;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final fun setupInitializationResponse ()V
+	public fun start (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+}
```
devcrocod, line 3386: I’m not sure why MockTransport needs to be added to the core module. It makes the core module more complex, and such things don’t really belong there.
In my opinion, there should be a separate module for mocks that’s used exclusively for testing

devcrocod (changes requested):
Please consider keeping `MockTransport` within the test code rather than exposing it in the public API

kotlin-sdk-core/api/kotlin-sdk-core.api
kpavlov: @devcrocod feel free to continue and move MockTransport to test classes and finish or drop this PR

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+ * The mock transport supports:
+ * - Recording all sent and received messages (via `getSentMessages` and `getReceivedMessages`)
+ * - Registering request handlers that respond to specific message predicates (e.g., by method)
+ * - Setting up responses that can be either successful or with errors
+ * - Waiting for specific messages to be received
+ *
+ * Note: This class is designed to be used as a test helper and should not be used in production.
+ */
+@Suppress("TooManyFunctions")
+public open class MockTransport(configurer: MockTransport.() -> Unit = {}) : Transport {
```
skarpovdev, line 40: why not use `AbstractTransport`?
if a test connects and sends immediately, then `onMessageBlock` may still be null if I understand correctly. and since `AbstractTransport` already has other safe handling, we may avoid other issues.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+        when (message) {
+            is JSONRPCRequest -> {
+                val response = requestHandlers.firstOrNull {
+                    it.first.invoke(message)
+                }?.second?.invoke(message)
+
+                checkNotNull(response) {
+                    "No request handler found for $message."
+                }
+                onMessageBlock?.invoke(response)
```
skarpovdev, line 76: should this one be wrapped in try/except and use `onError` in except clause?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/testing/MockTransport.kt
```diff
@@ -0,0 +1,213 @@
...
+     *
+     * @param poolInterval The interval at which the function polls the received messages. Default is 50 milliseconds.
+     * @param timeout The maximum time to wait for a matching message. Default is 3 seconds.
+     * @param timeoutMessage The error message to throw when the timeout is reached.
+     * Default is "No message received matching predicate".
+     * @param predicate A predicate function that returns true if the message matches the criteria.
+     * @return The first JSON-RPC message that matches the predicate.
+     */
+    @OptIn(ExperimentalTime::class)
+    public suspend fun awaitMessage(
```
skarpovdev, line 193: minor: I assume we can implement some signaling mechanism to avoid unnecessary delays.
like `withTimeout(timeout) { <some MutableSharedFlow>.first { predicate(it) } }`

kpavlov: Will other messages be lost in this case? The idea is to wait until the expected message has arrived, while preserving other messages, so they can be asserted too. It would work for concurrent scenario

e5l:
@kpavlov, could you check if we still need this PR?

kpavlov:
Let's redo it

PR #342, opened 2025-10-24
downgrade Ktor to 3.2.3

devcrocod:
- ktor 3.2.3
- Tests are ignored because Ktor 3.2.3 has issues with wasm in the test environment. However, it works outside of tests. All other targets pass these tests successfully.

## Motivation and Context
compatibility with older versions of Ktor

## How Has This Been Tested?
CI

## Checklist


devcrocod:
@tiginamaria Yes, I’m going to do that in the next PRs.
The WebSocket will be tested only with the JVM.

I’ll look into why sse fails with wasm and try to fix it for wasm. Otherwise, I’ll move it under the JVM as well.
At the moment, I didn’t want this to be a blocker for the release

Issue #295, opened 2025-09-28
Server should be accissible to tool handlers


rnett:
**Describe the bug**

Currently, there is no way to access the server instances in tool handlers. This is important to have so that tools can sent elicitation requests and other notifications.

Currently, the only workaround is to re-crease the tool handler for each server instance, which can get expensive.

rnett:
This is significantly worse with the introduction of server sessions.  A tool call handler has no way of knowing which session the call originated from, and thus no way of knowing where to send elicitations, notifications, etc.  This makes these features unusable.

@devcrocod should I create a new issue for this?

devcrocod:
@rnett Do you mean this issue: #249 ?

rnett:
No, I don't think so from that issue's description.  My issue is around the API, not implementation.  Inside of the handler you register a tool with (the `suspend (CallToolRequest) -> CallToolResult` lambda), there is no way to get the corresponding `ServerSession` instance to use to send notifications and messages back to the client.  Unless I'm missing something obvious?  I was able to work around this in previous versions by just using the server instance, since there was a single session, but that's not the case anymore, it's impossible to know which of the server's sessions a given tool invocation should send messages to.

rnett:
@devcrocod are there any updates here?  I see server session IDs have been added, but I don't see any way for a tool call handler to know its ID.  Are there tests or examples for a tool call sending messages back (e.g. logging or notifications) that i could look at to see that it works?  This is core functionality, so I would hope that it works.

devcrocod:
yeap
We do have plans to include the sessionId in `addTool` in one of the next releases

rnett:
Ok good, thanks. It's currently preventing me from updating, so it would be great if it was soon 🙂

rnett:
Any updates here?

rnett:
@devcrocod I'm happy to create a PR for this if it would speed things up.  I would plan to do it along the lines of https://github.com/modelcontextprotocol/kotlin-sdk/pull/144 (but with sessions, etc) unless I hear otherwise.

devcrocod:
hi @rnett 
Sorry for my long absence
I’ve looked at #144. The context-based approach looks like a reasonable solution, but it requires significant changes to the API. Additionally, I have some questions regarding the server-side implementation of such an API.
Could you please create a PR so we can discuss it in detail?

rnett:
[...]

PR #144, opened 2025-07-09
Add RegisteredTool receiver

mbiamont:
Add the `Server` as the receiver of the `RegisteredTool` handler lambda.
Same for `RegisteredPrompt` and `RegisteredResource`.

## Motivation and Context
- It's currently not easy to call methods like `sendLoggingMessage()` from a tool as you need an instance of the server.

Before:

```kotlin
// The server needs to be instantiated before any other tool
val server: Server = (...)

val registeredTool = RegisteredTool(
   tool = Tool(...),
)  {
   server.sendLoggingMessage(...)
}

server.addTools(listOf(registeredTool))
```

After:

```kotlin
// No need for Server instance
val registeredTool = RegisteredTool(
   tool = Tool(...),
)  {
  sendLoggingMessage(...)
}

val server: Server = (...).apply {
   addTools(listOf(registeredTool))
}
```

## How Has This Been Tested?
I've added a unit test.

## Breaking Changes
No breaking change

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

## Additional context

In the Python MCP SDK, a tool can have a `ctx: Context` parameter and this context has the methods for sending a notification message.

Example:

```python
from mcp.server.fastmcp import Context, FastMCP

mcp = FastMCP(name="Notifications Example")


@mcp.tool()
async def process_data(data: str, ctx: Context) -> str:
    """Process data with logging."""
    # Different log levels
    await ctx.debug(f"Debug: Processing '{data}'")
    await ctx.info("Info: Starting processing")
    await ctx.warning("Warning: This is experimental")
    await ctx.error("Error: (This is just a demo)")

    # Notify about resource changes
    await ctx.session.send_resource_list_changed()

    return f"Processed: {data}"
```

We may also want to not pass the whole `Server` instance as context of the lambda, but just a `Context` which could be defined as

```kotlin
interface McpServerContext {
   suspend fun sendLoggingMessage(params: LoggingMessageNotification)
   suspend fun sendResourceUpdated(params: ResourceUpdatedNotification)
[...]

Issue #236, opened 2025-08-25
Passing info about user/tenant to tools


ukarlsson:
**Describe the bug**

I can not understand how we should get information about the currently logged in user in the MCP request. We have added bearer authentication with OAuth flow for logging in to MCP, but it does not seem to be possible to pass information about the Ktor Principal or the tenant (can derive this from ApplicationCall). It seems to get lost here and I don't understand how to pass on more info here

```
    public suspend fun handlePostMessage(call: ApplicationCall) {
        if (!initialized.load()) {
            val message = "SSE connection not established"
            call.respondText(message, status = HttpStatusCode.InternalServerError)
            _onError.invoke(IllegalStateException(message))
        }

        val body = try {
            val ct = call.request.contentType()
            if (ct != ContentType.Application.Json) {
                error("Unsupported content-type: $ct")
            }

            call.receiveText()
        } catch (e: Exception) {
            call.respondText("Invalid message: ${e.message}", status = HttpStatusCode.BadRequest)
            _onError.invoke(e)
            return
        }

        try {
            handleMessage(body)
        } catch (e: Exception) {
            call.respondText("Error handling message $body: ${e.message}", status = HttpStatusCode.BadRequest)
            return
        }

        call.respondText("Accepted", status = HttpStatusCode.Accepted)
    }
```

So here we have ApplicationCall, but in handleMessage() it seems like we only have the body.

So we need a way to pass more arbitrary data. In particular:

- Ktor principal
- Any attributes we have on the ApplicationCall

So having full access to ApplicationCall downstream would be great!

Many thanks

kpavlov:
It would be great to access ApplicationCall, e.g., from ServerhandlerContext, see #515

PR #515, opened 2026-02-12
feat! Add context to server call handlers

rnett:
## Motivation and Context

Adds a `ClientConnection` context receiver to server call handlers, which provides access to the client communication channel. This is necessary to allow tools to send notifications, logging, elicitations, etc.

Supersedes https://github.com/modelcontextprotocol/kotlin-sdk/pull/144, discussed in https://github.com/modelcontextprotocol/kotlin-sdk/issues/295.

## How Has This Been Tested?

With the added integration test.

## Breaking Changes

It breaks binary compatibility and breaks source compatibility for tool handlers that already use `this`.

If you care a lot about maintaining compatibility, I could add some overloads to ease the transition.  But the new methods would likely need to be named differently for resolution.

Future compatibility should be easier to maintain - `Context` is an interface (with experimental opt-in subclassing), so new fields can be added without breaking compatibility.  Different types could also be introduced for different handles without breaking compatibility, as long as they extend `ClientConnection`.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/integration-test/detekt-baseline-test.xml
+++ b/integration-test/detekt-baseline-test.xml
@@ -32,7 +32,6 @@
     <ID>LongMethod:KotlinServerForTsClientSse.kt:HttpServerTransport$suspend fun handleRequest</ID>
     <ID>LongMethod:KotlinServerForTsClientSse.kt:KotlinServerForTsClient$fun createMcpServer: Server</ID>
     <ID>LongMethod:KotlinServerForTsClientSse.kt:KotlinServerForTsClient$fun start</ID>
-    <ID>LongMethod:KotlinTestBase.kt:KotlinTestBase$protected fun setupServer</ID>
     <ID>LongMethod:ServerResourcesNotificationSubscribeTest.kt:ServerResourcesNotificationSubscribeTest$@Test fun `should send resource notifications`</ID>
     <ID>LongMethod:TsEdgeCasesTestSse.kt:TsEdgeCasesTestSse$@Test @Timeout(30, unit = TimeUnit.SECONDS) fun testComplexConcurrentRequests: Unit</ID>
     <ID>MatchingDeclarationName:PromptIntegrationTestSse.kt:SchemaPromptIntegrationTestSse : AbstractPromptIntegrationTest</ID>
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/AbstractServerFeaturesTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/AbstractServerFeaturesTest.kt
@@ -5,9 +5,12 @@ import io.modelcontextprotocol.kotlin.sdk.client.ClientOptions
 import io.modelcontextprotocol.kotlin.sdk.shared.InMemoryTransport
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolResult
 import io.modelcontextprotocol.kotlin.sdk.types.ClientCapabilities
+import io.modelcontextprotocol.kotlin.sdk.types.GetPromptResult
 import io.modelcontextprotocol.kotlin.sdk.types.Implementation
+import io.modelcontextprotocol.kotlin.sdk.types.ReadResourceResult
 import io.modelcontextprotocol.kotlin.sdk.types.ServerCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.TextContent
+import io.modelcontextprotocol.kotlin.sdk.types.TextResourceContents
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import org.junit.jupiter.api.BeforeEach
@@ -21,6 +24,20 @@ abstract class AbstractServerFeaturesTest {
         server.addTool(name, "Test $name") {
             block()
             CallToolResult(listOf(TextContent("Success")))
+        }
+    }
+
+    protected fun addPrompt(name: String, block: suspend ClientConnection.() -> Unit) {
+        server.addPrompt(name, "Test $name") {
+            block()
+            GetPromptResult(messages = emptyList())
+        }
+    }
+
+    protected fun addResource(uri: String, block: suspend ClientConnection.() -> Unit) {
+        server.addResource(uri, uri, "Test resource $uri") {
+            block()
+            ReadResourceResult(contents = listOf(TextResourceContents(text = "content", uri = uri)))
         }
     }
 
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnectionTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnectionTest.kt
@@ -1,29 +1,40 @@
 package io.modelcontextprotocol.kotlin.sdk.server
 
-import io.kotest.matchers.collections.shouldHaveSize
+import io.kotest.assertions.assertSoftly
+import io.kotest.assertions.withClue
 import io.kotest.matchers.shouldBe
+import io.kotest.matchers.string.shouldNotBeBlank
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequest
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.ClientCapabilities
+import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequest
+import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.ListRootsRequest
 import io.modelcontextprotocol.kotlin.sdk.types.ListRootsResult
+import io.modelcontextprotocol.kotlin.sdk.types.LoggingLevel
+import io.modelcontextprotocol.kotlin.sdk.types.LoggingMessageNotification
+import io.modelcontextprotocol.kotlin.sdk.types.LoggingMessageNotificationParams
 import io.modelcontextprotocol.kotlin.sdk.types.Method
 import io.modelcontextprotocol.kotlin.sdk.types.Notification
 import io.modelcontextprotocol.kotlin.sdk.types.PromptListChangedNotification
+import io.modelcontextprotocol.kotlin.sdk.types.ReadResourceRequest
+import io.modelcontextprotocol.kotlin.sdk.types.ReadResourceRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.Request
 import io.modelcontextprotocol.kotlin.sdk.types.RequestResult
 import io.modelcontextprotocol.kotlin.sdk.types.ResourceListChangedNotification
+import io.modelcontextprotocol.kotlin.sdk.types.ResourceUpdatedNotification
+import io.modelcontextprotocol.kotlin.sdk.types.ResourceUpdatedNotificationParams
 import io.modelcontextprotocol.kotlin.sdk.types.Root
 import io.modelcontextprotocol.kotlin.sdk.types.ServerCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.ToolListChangedNotification
 import kotlinx.coroutines.CompletableDeferred
-import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.test.runTest
+import kotlinx.serialization.json.JsonPrimitive
 import org.junit.jupiter.api.Test
 
 class ClientConnectionTest : AbstractServerFeaturesTest() {
 
-    private inline fun <reified T : Notification> onClientNotification(method: Method): Deferred<T> {
+    private inline fun <reified T : Notification> onClientNotification(method: Method): CompletableDeferred<T> {
         val deferred = CompletableDeferred<T>()
         client.setNotificationHandler<T>(method) {
             deferred.complete(it)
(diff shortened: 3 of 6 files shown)


kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Context.kt
```diff
@@ -0,0 +1,14 @@
+package io.modelcontextprotocol.kotlin.sdk.server
+
+import io.modelcontextprotocol.kotlin.sdk.ExperimentalMcpApi
+
+/**
+ * Contextual information made available to server request handlers.
+ */
+@SubclassOptInRequired(ExperimentalMcpApi::class)
+public interface Context {
```
devcrocod: This name is too generic
to avoid conflicts, it would be better to use something more specific name

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Feature.kt
```diff
@@ -19,13 +19,15 @@ internal interface Feature {
     val key: FeatureKey
 }
 
+public typealias Handler<T, R> = suspend Context.(T) -> R
```
devcrocod: Also, in my opinion, this is not the best naming

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -193,7 +193,7 @@ public open class Server(
                 handleListTools()
             }
             session.setRequestHandler<CallToolRequest>(Method.Defined.ToolsCall) { request, _ ->
-                handleCallTool(request)
+                handleCallTool(session, request)
```
devcrocod, line 197: Could we also have methods specifically for cases where the session is provided externally?

devcrocod: I mean like that:
```
session.setRequestHandler<GetPromptRequest>(Method.Defined.PromptsGet) { request, _ ->
                handleGetPrompt(request)
            }
```
Because we are already calling this on a specific session

devcrocod (commented):
At the moment, I don’t quite see the need for a `Context` and just store `session`, if we only want to expose a session. We could achieve the same without requiring `session.someFun()` to be written every time

I’ll also try to find out whether it’s possible to overload functions by lambda. in the past, Kotlin had a special annotation for this. That could help us preserve compatibility.

I’m also concerned about the naming. Since this will be part of the public API, it would be better to use more specific names. Besides potential conflicts, this could also confuse users, as code completion will frequently suggest these symbols

cc: @tiginamaria

kpavlov:
I do not agree, that it is a kind of useless api. This is at least the second community PR: https://github.com/modelcontextprotocol/kotlin-sdk/pull/144

Context is pretty common pattern for server implementation and it provides is a good extension point, also could be used as a place to store some user attributes...
In software, API is never set in stone, unless it is sealed on a chip. We can annotate it `@ExperimentalMcpApi`.

rnett:
FWIW, the reason I went with `Context` over just session is that it allows for future changes and extensibility without breaking either source or binary compatibility.  Using session would for the current objectives, but if something changed in the future, it could easily cause another breaking change.

rnett:
RE overloads (@kpavlov), do you mean for the `registerTool`, etc., methods?  Because while agree it would be nice, it's not possible for Kotlin to resolve the overloads for a call like:
```kotlin
addTool(tool) {
  // is there a Context receiver or not? - the compiler reports an overload resolution error
}
```

The only alternative I could think of is using a different name for the new methods, like `register` instead of add.  Thoughts?

devcrocod:
> FWIW, the reason I went with Context over just session is that it allows for future changes and extensibility without breaking either source or binary compatibility. Using session would for the current objectives, but if something changed in the future, it could easily cause another breaking change.

Yes, that makes sense. What I meant was to attach extensions to the context in this case, so that the calls are made without a session. Since in the context we have only one variable, all calls will go through the session anyway, so specifying it every time is redundant

devcrocod:
> RE overloads (@kpavlov), do you mean for the registerTool, etc., methods? Because while agree it would be nice, it's not possible for Kotlin to resolve the overloads for a call like:

I checked and it looks like there’s only an annotation for overload resolution by return type `@OverloadResolutionByLambdaReturnType`.
Unfortunately, that won’t work in this case

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Context.kt
rnett: Updated to `ServerHandlerContext`.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Feature.kt
rnett: Inlined the typealias.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
rnett: I'm not entirely sure I get what you mean, but the only way I could see it working is if the `setRequestHandler`, etc, methods had a `Context` receiver.  Which would be nice, but doesn't work because they are defined in `Protocol`, which doesn't have enough information to create the context object.

The only way I could see it working is if `Protocol` had a "self" generic like `Protocol<T : Protocol>`, and `Context` was also made `Context<T : Protocol>`.  I can see the appeal, but I'm not sure if it's worth it, since someone setting the handler already has the session available.  Your call.

rnett:
> Yes, that makes sense. What I meant was to attach extensions to the context in this case, so that the calls are made without a session. Since in the context we have only one variable, all calls will go through the session anyway, so specifying it every time is redundant

Ah I see.  Personally, I don't find it that verbose, but I've updated the context so that you can call the methods directly on it.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
```diff
@@ -40,14 +40,103 @@ import kotlin.uuid.Uuid
 
 private val logger = KotlinLogging.logger {}
 
+/**
+ * Methods for communicating from the server back to the client.
+ */
+public interface ServerCommunication {
```
kpavlov: 1. Shouldn't we mark it as experimental today? I have a feeling that some methods should be a part of `Protocol`

```suggestion
@ExperimentalMcpApi
public interface ServerCommunication {
```

2. Should it be named as **ClientCommunication**, since it is exposing **client** API

3. Let's move it to separate file

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -294,7 +294,7 @@ public open class Server(
      * @param handler A suspend function that handles executing the tool when called by the client.
      * @throws IllegalStateException If the server does not support tools.
      */
-    public fun addTool(tool: Tool, handler: suspend (CallToolRequest) -> CallToolResult) {
+    public fun addTool(tool: Tool, handler: suspend ServerHandlerContext.(CallToolRequest) -> CallToolResult) {
```
kpavlov: ServerSession now implements both Protocol and ServerCommunication.  Our goal is to expose ServerCommunication, the client communication API interface, but we’re also implicitly exposing a server. This allows request handlers to be modified from within a tool handler call which I’m not particularly confident about.  Furthermore, it’s generally better to rely on interfaces rather than concrete classes.


```suggestion
    public fun addTool(tool: Tool, handler: suspend ServerCommunication.(CallToolRequest) -> CallToolResult) {
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -391,7 +391,10 @@ public open class Server(
      * @param promptProvider A suspend function that returns the prompt content when requested by the client.
      * @throws IllegalStateException If the server does not support prompts.
      */
-    public fun addPrompt(prompt: Prompt, promptProvider: suspend (GetPromptRequest) -> GetPromptResult) {
+    public fun addPrompt(
+        prompt: Prompt,
+        promptProvider: suspend ServerHandlerContext.(GetPromptRequest) -> GetPromptResult,
```
kpavlov: same as above
```suggestion
        promptProvider: suspend ServerCommunication.(GetPromptRequest) -> GetPromptResult,
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -412,7 +415,7 @@ public open class Server(
         name: String,
         description: String? = null,
         arguments: List<PromptArgument>? = null,
-        promptProvider: suspend (GetPromptRequest) -> GetPromptResult,
+        promptProvider: suspend ServerHandlerContext.(GetPromptRequest) -> GetPromptResult,
```
kpavlov: same as above
```suggestion
        promptProvider: suspend ServerCommunication.(GetPromptRequest) -> GetPromptResult,
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -479,7 +482,7 @@ public open class Server(
         name: String,
         description: String,
         mimeType: String = "text/html",
-        readHandler: suspend (ReadResourceRequest) -> ReadResourceResult,
+        readHandler: suspend ServerHandlerContext.(ReadResourceRequest) -> ReadResourceResult,
```
kpavlov: same as above
```suggestion
        readHandler: suspend ServerCommunication.(ReadResourceRequest) -> ReadResourceResult,
```

integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerHandlerNotificationTest.kt
kpavlov, line 1: Need tests to verify calling every method from ServerCommunication from within handlers. See all changed methods in Server

kpavlov:
maybe, something like https://github.com/modelcontextprotocol/kotlin-sdk/commit/35108b425ba06063428874d6d9f4a1333a82c23c
@devcrocod, WDYT?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
devcrocod: Can’t this be done using extension functions?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
devcrocod: 1. 
> Shouldn't we mark it as experimental today?

Yes, I think should

>  I have a feeling that some methods should be a part of Protocol

I’m not sure. In my opinion, we shouldn’t make changes to the protocol yet

2. In my opinion, it’s fine as it is. We can name it more generically and use the Mcp prefix
3. 👍

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
rnett: Are you proposing removing the context interface and using the communication interface as the receiver directly? That works, but means that if you ever want to add any additional context , you'll either need to make a breaking change to the receiver or add it to the communication interface. 

Also, if accessing the underlying server is a bit concern, we should wrap it in a class, currently you can just cast the receiver to server session.

kpavlov: > Are you proposing removing the context interface and using the communication interface as the receiver directly? That works, but means that if you ever want to add any additional context , you'll either need to make a breaking change to the receiver or add it to the communication interface.

True, but let's follow YAGNI principle. We can make this change later.  

> Also, if accessing the underlying server is a bit concern, we should wrap it in a class, currently you can just cast the receiver to server session.

That's true. One can cast easily. We should wrap it.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
rnett: For now, I've renamed it to `ClientConnection`, since it also contains information like the client version.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
rnett: Done.  I erred on the side of including too much, but I could easily tools and other handlers could wanting to know the client version, capabilities, etc.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
rnett: The only way I can see is by adding a `ClientConnection` receiver to `ServerSession.setRequestHandler` - is that what you mean?  I think it would make sense.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Feature.kt
```diff
@@ -35,9 +40,10 @@ public data class RegisteredTool(val tool: Tool, val handler: suspend (CallToolR
  * @property prompt The prompt definition.
  * @property messageProvider A suspend function that returns the prompt content when requested by the client.
  */
+@OptIn(ExperimentalMcpApi::class)
 public data class RegisteredPrompt(
     val prompt: Prompt,
-    val messageProvider: suspend (GetPromptRequest) -> GetPromptResult,
+    val messageProvider: suspend ClientConnection.(GetPromptRequest) -> GetPromptResult,
```
rnett, line 43: Thoughts on using a receiver vs a context parameter?  Receiver is slightly easier to access, but a context parameter (perhaps named `client`) might make usages more readable.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Feature.kt
```diff
@@ -25,7 +26,11 @@ internal interface Feature {
  * @property tool The tool definition.
  * @property handler A suspend function to handle the tool call requests.
  */
-public data class RegisteredTool(val tool: Tool, val handler: suspend (CallToolRequest) -> CallToolResult) : Feature {
+@OptIn(ExperimentalMcpApi::class)
```
rnett: Despite this (and similar opting-in in `Server`), actually using anything from `ClientConnection` still has a warning.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
```diff
@@ -590,50 +597,55 @@ public open class Server(
...
+        }
     }
 
     private fun handleListResourceTemplates(): ListResourceTemplatesResult {
         // If you have resource templates, return them here. For now, return empty.
         return ListResourceTemplatesResult(listOf())
     }
 
-    // Start the ServerSession redirection section
+    // Start the ServerSession / ClientConnection redirection section
```
rnett, line 637: Should we deprecate these?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
```diff
@@ -418,31 +308,50 @@ public open class ServerSession(
...
-
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other !is ServerSession) return false
         return sessionId == other.sessionId
     }
 
     override fun hashCode(): Int = sessionId.hashCode()
+
+    // Start the ClientConnection redirection section
```
rnett, line 320: I kept these because it would be a breaking change, and it's similar to what was done on `Server`.  LMK if it's not necessary.  Also, we could deprecate them.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ServerSession.kt
rnett: I think the name and pattern here better allows for things like https://github.com/modelcontextprotocol/kotlin-sdk/issues/236#issuecomment-3922127672, without having the extra "context" layer.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Feature.kt
rnett: Unfortunately, it also warns on every `addResource`, `addTool`, etc, method.  I tested various uses and opt-ins and was unable to get it not to.  Thoughts?  Marking every method of `ClientConnection` but not the class itself may work.

kpavlov (commented):
Thank you, @rnett 
Looks good, but I have couple of observations.

1. At Copilot mentioned, marking ClientConnection as experimental api requires OptIn for all addXxxHandler calls. This change will be disruptive for users and we do want to keep the current api shape, so it's not really experimental for the users. Let's remove ExperimentalMcpApi from ClientConnection

2. Maybe lets make ClientConnection an interface and create ClientConnectionImpl as internal class just for the sake of Dependency Inversion Principle, to have more flexibility in the future.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/Server.kt
kpavlov: probably

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -0,0 +1,270 @@
...
+import kotlinx.serialization.json.JsonObject
+
+private val logger = KotlinLogging.logger {}
+
+@Suppress("TooManyFunctions")
+public interface ClientConnection {
+    /**
+     * The capabilities supported by the server, related to the session.
+     */
+    public val serverCapabilities: ServerCapabilities
```
kpavlov: Should we expose it to the handler?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * The client's reported capabilities after initialization.
+     */
+    public val clientCapabilities: ClientCapabilities
```
kpavlov: Should we expose it to the handler?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -0,0 +1,270 @@
...
+    /**
+     * A unique identifier identifying the current session.
+     * Has no inherent meaning.
+     */
+    public val sessionId: String
+
+    /**
+     * Registers a callback to be invoked when the server session is closing.
+     */
+    public fun onClose(block: () -> Unit)
```
kpavlov: Should we expose it to the handler?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+     * Sends a request and waits for a response.
+     *
+     * Do not use this method to emit notifications! Use notification() instead.
+     */
+    public suspend fun <T : RequestResult> request(request: Request, options: RequestOptions? = null): T
```
kpavlov: Should we expose such low-level API to the handler?

kpavlov (dismissed):
LGTM

devcrocod (dismissed):
let’s merge it

rnett:
IMO some of the removed methods are quite useful and should be kept:
 * Knowing the client and server capabilities is necessary to gracefully handle either not supporting features. For example, to request an elicitation if supported, and fall back to something else (e.g. assume denied) if not. Without these, there is no way to determine if the message methods will throw or not. Version could be useful for similar reasons.
 * onClose is useful to clean up long running or background jobs in once a client disconnects. For example, my MCP server has a tool to run a Gradle task in the background, and I would like to terminate any tasks started by a session when it is closed. 
 * The logging check messages are useful to avoid computing and processing log messages for levels that will be ignored, the same as the `logger.debug { message }` apis. 
 * If you remove request, I think it makes sense to remove notification, too. I forget whether the hierarchies are open or not - the main reason I included them is to handle a case where a notification or message type exists in the hierarchy, but doesn't yet have a dedicated method.

kpavlov:
Let's reconsider each method separately, one at a time with a comprehensive testing added to each case in the individual PR. 

All those mentioned do not fit into such huge PR like this. It took quite some time to add missing tests.
Let's respect reviewer's time and mental efforts.

rnett:
👍 I'll create issues for the others

PR #516, opened 2026-02-12
refactor: Introduce AbstractClientTransport and simplify client transport lifecycle (#517)

kpavlov:
## refactor: Introduce AbstractClientTransport and simplify client transport lifecycle

This PR unifies state management logic across client transports, except for the websocket transport, as it needs more refactoring first and is not the primary transport.

- Introduce `AbstractClientTransport` as base class for other client transports.
- Migrate to `ClientTransportState` states with state machine, streamline state management.
- Ensure proper cleanup and error handling consistency across transport implementations.
- Add unit tests to validate `AbstractClientTransport` behavior and state transitions.
- Update related documentation and API contract.
- Make sure `onClose` callback is called at most once
- fix [StreamableHttpClientTransportTest.kt‎](https://github.com/modelcontextprotocol/kotlin-sdk/pull/516/changes#diff-b9fe8891d08a492448a7ed2dc782151c6a915ae1403a4d70c2034c6062c84cb5) to throw `McpException` on connect when received invalid Json

## Motivation and Context

Unify lifecycle handling and eliminate duplication in client Transport implementations.

## How Has This Been Tested?
Unit test, regression tests

## Breaking Changes

- `onClose` callback is now called only once
- send(...) will throw McpException with message "Transport is not ready" when `ClientTransportState != Operational`

## Types of changes

- [x] Refactoring
- [ ] Documentation update

## Checklist

## Additional context
See #517

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
@@ -1,5 +1,6 @@
 package io.modelcontextprotocol.kotlin.sdk.client
 
+import io.github.oshai.kotlinlogging.KLogger
 import io.github.oshai.kotlinlogging.KotlinLogging
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.sse.ClientSSESession
@@ -45,9 +46,7 @@ public class SseClientTransport(
     private val requestBuilder: HttpRequestBuilder.() -> Unit = {},
 ) : AbstractClientTransport() {
 
-    private companion object {
-        private val logger = KotlinLogging.logger {}
-    }
+    override val logger: KLogger = KotlinLogging.logger {}
 
     private val endpoint = CompletableDeferred<String>()
 
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
@@ -1,6 +1,7 @@
 package io.modelcontextprotocol.kotlin.sdk.shared
 
-import io.github.oshai.kotlinlogging.KotlinLogging
+import io.github.oshai.kotlinlogging.KLogger
+import io.modelcontextprotocol.kotlin.sdk.InternalMcpApi
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCMessage
 import io.modelcontextprotocol.kotlin.sdk.types.McpException
 import io.modelcontextprotocol.kotlin.sdk.types.RPCError.ErrorCode.CONNECTION_CLOSED
@@ -20,18 +21,46 @@ import kotlin.coroutines.cancellation.CancellationException
  * to ensure invariants during lifecycle events. The initialization and shutdown logic is designed to handle
  * async workflows, ensuring proper resource cleanup in the case of failure or cancellation.
  */
-@OptIn(ExperimentalAtomicApi::class)
+@OptIn(ExperimentalAtomicApi::class, InternalMcpApi::class)
 public abstract class AbstractClientTransport : AbstractTransport() {
 
-    private companion object {
-        private val logger = KotlinLogging.logger {}
-    }
+    protected abstract val logger: KLogger
+
+    /**
+     * Represents the current state of the client transport in an atomic and thread-safe manner.
+     *
+     * This private field is used internally to manage state transitions of the transport, ensuring
+     * consistency across concurrent operations. The state transitions are validated and managed
+     * using the `stateTransition` method, adhering to defined rules in [ClientTransportState.VALID_TRANSITIONS].
+     *
+     * @see ClientTransportState
+     * @see stateTransition
+     * @see checkState
+     */
+    private val _state: AtomicReference<ClientTransportState> = AtomicReference(ClientTransportState.New)
 
     /**
      * Current transport state. Thread-safe via atomic reference.
      * Initial state is [ClientTransportState.New].
      */
-    protected val state: AtomicReference<ClientTransportState> = AtomicReference(ClientTransportState.New)
+    @InternalMcpApi
+    protected val state: ClientTransportState
+        get() = _state.load()
+
+    /**
+     * Updates the internal state of the client transport to the specified new state.
+     *
+     * This method is used to manage the lifecycle and operational state of the client transport
+     * by transitioning it to the provided [new] state. The new state is stored atomically,
+     * ensuring thread safety during concurrent modifications.
+     *
+     * @param new The new state to transition the client transport to. It must be a valid state
+     *            as defined in the transport's lifecycle logic.
+     */
+    @InternalMcpApi
+    protected fun updateState(new: ClientTransportState) {
+        _state.store(new)
+    }
 
     /**
      * Performs transport-specific initialization (connections, I/O streams, background coroutines).
@@ -46,14 +75,15 @@ public abstract class AbstractClientTransport : AbstractTransport() {
      *
      * Validates transition per [ClientTransportState.VALID_TRANSITIONS] and performs atomic compare-and-exchange.
      *
-     * @throws IllegalArgumentException if transition not allowed
-     * @throws IllegalStateException if actual state doesn't match `from`
+     * @throws IllegalArgumentException if transition isn't allowed
+     * @throws IllegalStateException if the actual state doesn't match `from`
      */
+    @InternalMcpApi
     protected fun stateTransition(from: ClientTransportState, to: ClientTransportState) {
         require(to in ClientTransportState.VALID_TRANSITIONS.getValue(from)) {
             "Invalid transition: $from → $to"
         }
-        val actualState = state.compareAndExchange(
+        val actualState = _state.compareAndExchange(
             expectedValue = from,
             newValue = to,
         )
@@ -78,7 +108,7 @@ public abstract class AbstractClientTransport : AbstractTransport() {
             "Expected transport state $expected, but found $it"
         },
     ) {
-        val actualState = state.load()
+        val actualState = state
         check(actualState == expected) { lazyMessage }
     }
 
@@ -92,15 +122,20 @@ public abstract class AbstractClientTransport : AbstractTransport() {
      *
      * This method operates within a suspend context to allow for asynchronous initialization tasks.
      *
-     * Throws exception if the initialization process fails and transfer state to [ClientTransportState.InitializationFailed].
+     *  **Important! This method will eventually become `final`.
+     *  Please don't override it, or consistency guarantees might be lost. Override [initialize] instead.**
+     *
+     * @throws Exception if the initialization process fails and transfers state
+     *                  to [ClientTransportState.InitializationFailed].
+     * @see initialize
      */
-    public final override suspend fun start() {
+    public override suspend fun start() {
         stateTransition(from = ClientTransportState.New, to = ClientTransportState.Initializing)
         try {
             initialize()
             stateTransition(from = ClientTransportState.Initializing, to = ClientTransportState.Operational)
         } catch (e: Exception) {
-            state.store(ClientTransportState.InitializationFailed)
+            _state.store(ClientTransportState.InitializationFailed)
             closeResources()
             throw e
         }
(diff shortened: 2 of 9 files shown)


kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
```diff
@@ -44,10 +43,12 @@ public class SseClientTransport(
     private val urlString: String?,
     private val reconnectionTime: Duration? = null,
     private val requestBuilder: HttpRequestBuilder.() -> Unit = {},
-) : AbstractTransport() {
-    private val logger = KotlinLogging.logger {}
+) : AbstractClientTransport() {
+
+    private companion object {
```
devcrocod: Why do we need a companion object?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Atomically transitions state from `from` to `to`.
+     *
+     * Validates transition per [ClientTransportState.VALID_TRANSITIONS] and performs atomic compare-and-exchange.
+     *
+     * @throws IllegalArgumentException if transition not allowed
+     * @throws IllegalStateException if actual state doesn't match `from`
+     */
+    protected fun stateTransition(from: ClientTransportState, to: ClientTransportState) {
```
devcrocod, line 82: I’m confused by that `stateTransition` and `checkState` are protected, while `start`, `send` and `close` are final, and it’s unclear where those functions are actually used. In that case, maybe they should be private instead?

devcrocod: or, opposite, make it public

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * Releases transport-specific resources (coroutines, connections, streams).
+     *
+     * Called by [close] during [ClientTransportState.SHUTTING_DOWN].
+     * May be called even if [initialize] failed partially.
+     * **DO NOT** call [_onClose] from this method - [close] handles it in the finally block.
+     */
+    protected open suspend fun closeResources() {
```
devcrocod: in my opinion, it would be better to make this function abstract
otherwise, it’s easy to miss that it needs to be overridden, which could eventually lead to a leak

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
```diff
@@ -51,4 +54,18 @@ public abstract class AbstractTransport : Transport {
...
+    /**
+     * Invokes the `_onClose` callback if it has not been already triggered.
+     *
+     * This method ensures the `_onClose` callback is executed only once by utilizing
+     * an atomic flag (`onCloseCalled`). If the callback has already been executed,
+     * the method does nothing. Any exceptions thrown during the execution of the
+     * `_onClose` callback are caught and suppressed.
+     */
+    protected fun invokeOnCloseCallback() {
+        if (onCloseCalled.compareAndSet(expectedValue = false, newValue = true)) {
```
devcrocod, line 67: Why do we need a flag and a check here?
The fact that `_onClose` is called is already controlled at the transport level. Also, can `_onClose` be called more than once?
In that case, on the second call `_onClose` would never be executed

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ClientTransportState.kt
```diff
@@ -0,0 +1,79 @@
...
+ * defined under the MCP [Protocol Lifecycle](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle).
+ */
+public enum class ClientTransportState {
+
+    /**
+     * The transport has just been created.
+     *
+     * This is the Initial state
+     */
+    NEW,
```
devcrocod: Please check that our Enum naming style is consistent across the project. As far as I remember, we also used UpperCamelCase in some places

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
```diff
@@ -51,4 +54,18 @@ public abstract class AbstractTransport : Transport {
...
+
+    /**
+     * Invokes the `_onClose` callback if it has not been already triggered.
+     *
+     * This method ensures the `_onClose` callback is executed only once by utilizing
+     * an atomic flag (`onCloseCalled`). If the callback has already been executed,
+     * the method does nothing. Any exceptions thrown during the execution of the
+     * `_onClose` callback are caught and suppressed.
+     */
+    protected fun invokeOnCloseCallback() {
```
devcrocod, line 66: The naming is inconsistent with the other `_onMessage` and `_onError`

devcrocod (changes requested):
good PR
I have a few clarifying questions, and there is also a failure with the conformance test

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
kpavlov: Check state/transition should only be accessed from subclasses so they are protected. Internal api should never be public!

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
kpavlov: It's a common pattern for loggers to live in companion objects so they are created when the class is loaded.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
kpavlov: Make sense. I can change it in a follow up PR

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
kpavlov: Controlling it on a transport level is causing code duplication, since such flag should be on every subclasses. This is the fix for duplicated logic. Open-closed principle

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
kpavlov: It's not a banking property, it's protected method. Naming is correct

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ClientTransportState.kt
kpavlov: Ok, I can change it for consistency.
But I am courious why CamelCase naming was choosen in the project. Even if Kotlin naming convention allows it, the common practice in JVM is to have UPPER_CASE names for enums, including examples in Kotlin documentation.
As it is not enforced by the linter - it is valid. I can change it in the follow-up pr

kpavlov:
@devcrocod, do you see any blockers? Or this PR, even if not perfect, makes the code better?

Conformance test is flaky and it is passed on the second attempt.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
devcrocod: This is a private object, so it can be left as a property. This will simplify the code. But it’s not a problem

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
devcrocod: Okay, but then why make start, send, and close final? This prevents users from overriding these methods to suit their own needs
And this could also be a breaking change if users already have these methods overridden

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
devcrocod: It was used only in a few transports, not at all. In my view, TransportState addresses this issue.

This also doesn’t eliminate duplication, since a different flag was used before, which was toggled from true to false.

And it doesn’t answer the question of multiple _onClose calls during the lifecycle of a Transport

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
devcrocod: I’m not saying naming is wrong
I’m saying it’s inconsistent, since in the code we have both _onMessage() and invokeOnCloseCallback

devcrocod:
@kpavlov Yes, I think there are two important questions:
- https://github.com/modelcontextprotocol/kotlin-sdk/pull/516#discussion_r2801293833
- https://github.com/modelcontextprotocol/kotlin-sdk/pull/516#discussion_r2801304841

Since this could potentially affect both compatibility and behavior

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
kpavlov: This is classical dilemma with Template Method pattern, when we want to balance consistency and flexibility. If users override those methods, we can't guarantee consistency. That's why template methods are usually final.

This is breaking change indeed, and users are forced to override different methods or they have to extend their transport from AbstractTransport or Transport interface.

I agree to make methods final in 0.10 to minimize disruption.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
kpavlov: Regarding multiple calls to _onClose : it simply doesn't make sense. This change is a **bugfix.**

Obviously, once closed, the transport should not be reopened. Or it should be new callback, e.g. "onDisconnect". Calling _onClose twice confuses user's logic: why are we closing already closed transport. Users would have to implement idempotency on their side.

Regarding duplication: It is eliminated, as `val onCloseCalled: AtomicBoolean` was removed in subclasses.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
kpavlov: Let's rename _onMessage(), but in a separate PR

kpavlov:
@devcrocod, I have addressed some questions you have rised, thank you.

> * [refactor: Introduce AbstractClientTransport and simplify client transport lifecycle (#517) #516 (comment)](https://github.com/modelcontextprotocol/kotlin-sdk/pull/516#discussion_r2801293833)

Addressed

> * [refactor: Introduce AbstractClientTransport and simplify client transport lifecycle (#517) #516 (comment)](https://github.com/modelcontextprotocol/kotlin-sdk/pull/516#discussion_r2801304841)

I have addressed the question in [this pull request](https://github.com/modelcontextprotocol/kotlin-sdk/pull/516#discussion_r2802487301). I have also removed one more leftover item. However, I will not revert the bugfix.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport.kt
devcrocod: Yes, maybe earlier this was an issue
I mean that previously `_onClose` depended on the state in the transport. Now there is TransportState and a separate state for _onClose, and they are independent. This can potentially lead to a bug if these states diverge

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/AbstractClientTransport.kt
devcrocod: Yes, I understand the logic. What confused me is that `stateTransition` is protected but is not used in subclasses and is used exclusively in final methods. Accordingly, I would expect `stateTransition` to be private, or for `start`, `send` and `close` to be open for overriding

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ClientTransportState.kt
devcrocod: As far as I remember, we took this from ktor, where CamelCase is used

devcrocod (approved):
👍

PR #530, opened 2026-02-17
feat(dsl): add result DSL builders

kpavlov:
# Add result DSL builders and tests

- Introduced `ToolsResultDslTest` and `PromptsResultDslTest` to test result builders for tool and prompt operations.
- Added `buildCompleteResult` and `buildInitializeResult` for server-side result creation.
- Updated API surface to include result DSL functionality.
- Added new test cases for `CallToolResult`, `ListResourcesResult`, and `ReadResourceResult` to validate error handling, structured content, and metadata processing.
- Improved coverage for resource and template handling with additional edge cases for mixed-field resources and pagination
- Updated integration tests to use the `buildCallToolResult` DSL for constructing results, improving clarity and reducing boilerplate. Added `@OptIn(ExperimentalMcpApi::class)` annotations to test files requiring experimental API.
- Refactored delay handling in tests to streamline coroutine usage.

## Motivation and Context

Build also results with DSL

## How Has This Been Tested?
Unit test

## Breaking Changes
No

## Types of changes

- [x] Tests update
- [ ] Documentation update

## Checklist

### DIFF
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractToolIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractToolIntegrationTest.kt
@@ -1,6 +1,7 @@
 package io.modelcontextprotocol.kotlin.sdk.integration.kotlin
 
 import io.kotest.assertions.json.shouldEqualJson
+import io.modelcontextprotocol.kotlin.sdk.ExperimentalMcpApi
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequest
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolResult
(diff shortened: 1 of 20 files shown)


e5l (commented):
The only thing I'm not sure about is having DSL functions on the top level. Users will receive the top-level functions `buildListResourcesResult` for the whole application. While it is fine for stdlib, doing this in other library will bloat top-level scope.

Should we think about scoping the DSL to specific use-sites or anchor classes?

@devcrocod, what do you think?

kpavlov:
> Should we think about scoping the DSL to specific use-sites or anchor classes?

Agree
But it might require merging #515 first where we introduce handler context object as a receiver. It will be a perfect place, but I am not sure about the visibility yet, as server handler context lives in the server module.

What if we try addressing it in the follow up PR, after #515?

kpavlov:
Actually, I'm not sure we should be very restrictive in letting users to build data classes

devcrocod:
> The only thing I'm not sure about is having DSL functions on the top level. Users will receive the top-level functions buildListResourcesResult for the whole application. While it is fine for stdlib, doing this in other library will bloat top-level scope.

That’s true. Even in the stdlib this pattern is used rarely. Given that we already have quite a few objects, relying on a shared prefix would indeed make the api awkward to use

> Should we think about scoping the DSL to specific use-sites or anchor classes?

I’ve thought about that as well, but introducing contextual scoping would require a more detailed design

devcrocod:
Closing this PR as the design discussion around top-level DSL scoping remains unresolved, and the branch has diverged significantly from `main`

PR #535, opened 2026-02-18
feat: Verify MCP endpoints with authentication for Streamable HTTP transport, test for SSE transport

kpavlov:
# Verify/fix MCP endpoint with Ktor authentication for SSE and Streamable-HTTP

1. Add Ktor authentication integration tests for protecting MCP routes  (SSE & Streamable HTTP)
2. Add/refactor MCP route extensions 
    - Add missing Route.mcpStreamableHttp extensions, similar to other transports
    - Move `TransportManager` to a separate file and generalize.
    - Refactored `mcpStreamableHttp` and `mcpStatelessStreamableHttp` extensions to use the config object instead of too many parameters. This change unblocks #521
    - Added integration tests (`KtorStreamableHttpExtensionsTest`) to validate route registrations, subpath handling, and sibling route isolation.

## Motivation and Context

To demonstrate and verify Ktor authentication in action, I provided an example. However, it revealed that StreamableHttp transport isn’t fully supported. Consequently, missing extensions were implemented and method signatures were refactored.

## How Has This Been Tested?
Unit and integration tests

## Breaking Changes

Only the new API was modified.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context

#521

Issue #521, opened 2026-02-12
Support configuration of the max request payload size for Streamable HTTP impl


mwaitt-wd:
**Is your feature request related to a problem? Please describe.**
There is currently a hardcoded [request body limit](https://github.com/modelcontextprotocol/kotlin-sdk/blob/main/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt#L44) used in the `StreamableHttpServerTransport`, set at 4 MB. We want to set this lower in our server, and it's a bit of a pain to implement this currently.

**Describe the solution you'd like**
It would be great if this were configurable via a function or parameter on the consumer side. Even if the SDK needed to enforce a "hard" 4 MB limit on that configured maximum, it would just be nice to have the ability to set one that is lower than that.

**Describe alternatives you've considered**
We are working on an approach to count the request payload size and assert that it's below our lower threshold manually, before passing it to the SDK. Afaict there really isn't an alternative to this at the moment.

kpavlov:
Yes, it is definitely a good idea to have these parameters configurable.

jskjw157:
@kpavlov I'd like to work on this. I'm planning to add a `maxRequestBodySize` parameter to the existing `StreamableHttpServerTransport.Configuration` class, defaulting to the current 4 MB limit. This will allow consumers to set a lower threshold without manual payload counting.

kpavlov:
Thank you, @jskjw157 
Yes, `StreamableHttpServerTransport.Configuration` is the right place for such parameter!

Issue #249, opened 2025-09-10
Fix notification pipeline


tiginamaria:
**Describe the bug**
Notification on subscription and on resources/tools/prompts change does not work for now

**To Reproduce**
1. 

**Expected behavior**

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

PR #560, opened 2026-02-23
feat: introduce StreamableHttpServerTransport.Configuration

kpavlov:
## Introduce Configuration class for StreamableHttpServerTransport

Replace the six-parameter flat constructor of `StreamableHttpServerTransport` with a typed `Configuration` class. This improves API ergonomics, enables structural equality and copy(), and provides a stable extension point for future options without further breaking the constructor signature.

### Changes:

  - Add `Configuration` as a public class nested directly on `StreamableHttpServerTransport`, with `enableJsonResponse` as the first parameter (most commonly set)
  - Change the primary constructor to accept `Configuration`.
  - Rename `retryIntervalMillis: Long?` to `retryInterval: Duration?` in Configuration, aligning with Kotlin's type-safe time API
  - Deprecate the old flat constructor with a compatibility bridge
  - Update KotlinTestBase integration test to use the new constructor
  - Enable test AbstractResourceIntegrationTest.testSubscribeAndUnsubscribe() since #249 is closed

## Motivation and Context

The current StreamableHttpServerTransport API cannot be easily extended: adding more parameters would be a breaking change. But this is already needed for #521. 
This PR is a prerequisite for #535

## How Has This Been Tested?
Regression tests

## Breaking Changes
No. Current StreamableHttpServerTransport constructor was deprecated

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
@@ -74,6 +74,9 @@ private data class SessionContext(val session: ServerSSESession?, val call: Appl
 @Suppress("TooManyFunctions")
 public class StreamableHttpServerTransport(private val configuration: Configuration) : AbstractTransport() {
 
+    @Deprecated("Use default constructor with explicit Configuration()")
+    public constructor() : this(configuration = Configuration())
+
     /**
      * Secondary constructor for `StreamableHttpServerTransport` that simplifies initialization by directly taking the
      * configurable parameters without requiring a `Configuration` instance.
@@ -91,9 +94,12 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
      * @param retryIntervalMillis Retry interval in milliseconds for event handling or reconnection attempts.
      *          Defaults to `null`.
      */
+    @Suppress("MaxLineLength")
     @Deprecated(
         "Use constructor with Configuration: StreamableHttpServerTransport(Configuration(enableJsonResponse = ...))",
-        level = DeprecationLevel.WARNING,
+        replaceWith = ReplaceWith(
+            "StreamableHttpServerTransport(Configuration(enableJsonResponse = enableJsonResponse, enableDnsRebindingProtection = enableDnsRebindingProtection, allowedHosts = allowedHosts, allowedOrigins = allowedOrigins, eventStore = eventStore, retryIntervalMillis = retryIntervalMillis))",
+        ),
     )
     public constructor(
         enableJsonResponse: Boolean = false,
--- a/kotlin-sdk-server/api/kotlin-sdk-server.api
+++ b/kotlin-sdk-server/api/kotlin-sdk-server.api
@@ -194,6 +194,7 @@ public final class io/modelcontextprotocol/kotlin/sdk/server/StdioServerTranspor
 
 public final class io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport : io/modelcontextprotocol/kotlin/sdk/shared/AbstractTransport {
 	public static final field STANDALONE_SSE_STREAM_ID Ljava/lang/String;
+	public fun <init> ()V
 	public fun <init> (Lio/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport$Configuration;)V
 	public fun <init> (ZZLjava/util/List;Ljava/util/List;Lio/modelcontextprotocol/kotlin/sdk/server/EventStore;Ljava/lang/Long;)V
 	public synthetic fun <init> (ZZLjava/util/List;Ljava/util/List;Lio/modelcontextprotocol/kotlin/sdk/server/EventStore;Ljava/lang/Long;ILkotlin/jvm/internal/DefaultConstructorMarker;)V


integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractResourceIntegrationTest.kt
```diff
@@ -165,7 +165,6 @@ abstract class AbstractResourceIntegrationTest : KotlinTestBase() {
         assertEquals(testResourceContent, content.text, "Resource content should match")
     }
 
-    @Ignore("Blocked by https://github.com/modelcontextprotocol/kotlin-sdk/issues/249")
```
kpavlov, line 168: This test is passing now

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -66,32 +68,86 @@ private data class SessionContext(val session: ServerSSESession?, val call: Appl
...
+        public val enableJsonResponse: Boolean = false,
+        public val enableDnsRebindingProtection: Boolean = false,
+        public val allowedHosts: List<String>? = null,
+        public val allowedOrigins: List<String>? = null,
+        public val eventStore: EventStore? = null,
+        public val retryInterval: Duration? = null,
+    ) {
+        public companion object {
+            public val Default: Configuration = Configuration()
+        }
```
devcrocod: Why is a companion object needed?
In my view, it’s redundant, considering that Configuration already has all the parameters with default values

devcrocod (commented):
Except for one place where I left a comment, it looks good

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
kpavlov: It serves as a constant avoid creating Configuration instances every time. Does it make sense?

devcrocod: Not really.
Is it actually expected that users will be creating thousands of Configuration instances?
Also, this adds it to the public Companion

kpavlov: > Is it actually expected that users will be creating thousands of Configuration instances?


Thank you, @devcrocod.
When I took a closer look at the code, I noticed a more important problem [here](https://github.com/modelcontextprotocol/kotlin-sdk/blob/f649ff0a7e3d258329d2dcb9dc62a2b05224cc39/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt#L162): the custom config object is being created **on every POST request**, anda one-time StreamableHttpServerTransport is also being created.

The Configuration object would have been reused if [this method](https://github.com/modelcontextprotocol/kotlin-sdk/blob/f649ff0a7e3d258329d2dcb9dc62a2b05224cc39/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt#L149-L155) had accepted it as a parameter.

Given this, let’s remove the default parameter value from the StreamableHttpServerTransport constructor completely. So, StreamableHttpServerTransport.Default is also not needed.

Changing `Application.mcpStatelessStreamableHttp(...)` parameters from plain parameter list to configuration object should be separate PR.

kpavlov:
Please, have a look at #563 after this one

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -66,32 +68,81 @@ private data class SessionContext(val session: ServerSSESession?, val call: Appl
...
- * @param eventStore Event store for resumability support
- *          If provided, resumability will be enabled, allowing clients to reconnect and resume messages
- * @param retryIntervalMillis Retry interval (in milliseconds) advertised via SSE priming events
- *          to hint the client when to reconnect. Applies only when an [eventStore] is configured.
- *          Defaults to `null` (no retry hint).
+ * @param configuration Transport configuration. See [Configuration] for available options.
  */
 @OptIn(ExperimentalUuidApi::class, ExperimentalAtomicApi::class)
 @Suppress("TooManyFunctions")
-public class StreamableHttpServerTransport(
```
e5l, line 87: Should we add a method for backward source compatibility?

e5l (dismissed):
lgtm

aozherelyeva (dismissed):

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
kpavlov: done

Issue #329, opened 2025-10-22
modelscope : Error connecting to transport: Error POSTing to endpoint


liuyuesheng:
**Describe the bug**
ask modelscope mcp is error
url:https://mcp.api-inference.modelscope.net/be66a2296a924f/sse

and because the baseUrl is Error
debug baseUrl is **https://mcp.api-inference.modelscope.net/be66a2296a924f**

<img width="1266" height="377" alt="Image" src="https://github.com/user-attachments/assets/8da244e4-b4e2-4ddb-a045-ddf2278700c8" />

but the real baseUrl is  **https://mcp.api-inference.modelscope.net/**

and this url is ok  in ts sdk

i think kotlin sdk need consistent with other sdk

liuyuesheng:
Temporary solution:

val baseUrlField = transport.javaClass.getDeclaredField("baseUrl$delegate")
baseUrlField.isAccessible = true
val baseUrl = baseUrlField.get(transport)
val _value = baseUrl.javaClass.getDeclaredField("_value")

_value.isAccessible = true
_value.set(baseUrl,"https://mcp.api-inference.modelscope.net/")

jabin-ma:
The implementation of handleEndpoint is incorrect. When the endpoint starts with a forward slash /, we should concatenate it with protocolWithAuthority instead of the full baseUrl. Here is the refined implementation:
```
private val hostUrl: String by lazy {
        session.call.request.url.protocolWithAuthority
}

val prefix = if (eventData.startsWith("/")) hostUrl else baseUrl
val endpointUrl = Url("$prefix/$eventData")
```

Amaneusz:
Hey @liuyuesheng, thank you for submitting this issue and sorry for answering with delay.
We have verified and fixed this bug—the fix should be available in the next release.

PR #565, opened 2026-02-26
fix: fix SseTransport endpoint resolution (#329)

Amaneusz:
## Summary
Fix SSE endpoint resolution so absolute paths resolve against the origin, and add tests covering absolute vs relative endpoint events.

## Motivation and Context
The SSE transport currently resolves all endpoint events against the SSE URL’s base path. When the server sends an absolute endpoint (starting with `/`), the client incorrectly prefixes it with the SSE path. This change resolves absolute endpoints against the origin while keeping relative behavior unchanged.

addresses: https://github.com/modelcontextprotocol/kotlin-sdk/issues/329

## How Has This Been Tested?
Unit tests added in `SseClientTransportTest` for absolute and relative endpoint events.

## Breaking Changes
No.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
The endpoint resolution logic now treats paths starting with `/` as origin-relative. Relative paths continue to resolve against the SSE base URL.

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
@@ -11,7 +11,6 @@ import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.ContentType
 import io.ktor.http.HttpHeaders
-import io.ktor.http.Url
 import io.ktor.http.append
 import io.ktor.http.isSuccess
 import io.ktor.http.protocolWithAuthority
@@ -54,13 +53,17 @@ public class SseClientTransport(
     private lateinit var scope: CoroutineScope
     private var job: Job? = null
 
+    private val origin: String by lazy {
+        session.call.request.url.protocolWithAuthority
+    }
+
     private val baseUrl: String by lazy {
         session.call.request.url.let { url ->
             val path = url.encodedPath
             when {
-                path.isEmpty() -> url.protocolWithAuthority
-                path.endsWith("/") -> url.protocolWithAuthority + path.removeSuffix("/")
-                else -> url.protocolWithAuthority + path.take(path.lastIndexOf("/"))
+                path.isEmpty() -> origin
+                path.endsWith("/") -> origin + path.removeSuffix("/")
+                else -> origin + path.take(path.lastIndexOf("/"))
             }
         }
     }
@@ -136,13 +139,16 @@ public class SseClientTransport(
     }
 
     /**
-     * Resolves and completes [endpoint] based on [eventData]
-     * Treats absolute paths as relative to origin and relative paths as relative to [baseUrl]
+     * Resolves and completes [endpoint] based on [eventData].
+     * Uses full URLs as-is, treats absolute paths as origin-relative,
+     * and relative paths as relative to [baseUrl].
      */
     private fun handleEndpoint(eventData: String) {
         try {
-            val endpointUrl = if (eventData.startsWith("/")) {
-                Url(baseUrl).protocolWithAuthority + eventData
+            val endpointUrl = if (eventData.startsWith("http://") || eventData.startsWith("https://")) {
+                eventData
+            } else if (eventData.startsWith("/")) {
+                origin + eventData
             } else {
                 "$baseUrl/$eventData"
             }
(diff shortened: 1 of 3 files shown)


kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/SseClientTransportTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+class SseClientTransportTest {
+
+    @Test
+    fun `absolute path endpoint resolves against origin`() = runTest {
+        // Given
+        val sseUrl = "http://example.com/api/mcp/sse"
+
+        // And
+        val endpointEvent = "/messages?sessionId=abc"
```
kpavlov, line 21: Let's also add a usecase when server emits a full URL such as https://example.com/messages?sessionId=abc. The [MCP protocol specfication](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse) doesn't restrict it

The current code would treat this as a relative path and produce `http://example.com/api/mcp/https://example.com/messages?sessionId=abc`, which is wrong. 

The endpoint url resolution might look like:
```kotlin
val endpointUrl = when {
      eventData.startsWith("http://") || eventData.startsWith("https://") -> eventData
      eventData.startsWith("/") -> origin + eventData
      else -> "$baseUrl/$eventData"
  }
```

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
```diff
@@ -135,11 +135,18 @@ public class SseClientTransport(
...
         try {
-            val path = if (eventData.startsWith("/")) eventData.substring(1) else eventData
-            val endpointUrl = Url("$baseUrl/$path")
-            endpoint.complete(endpointUrl.toString())
+            val endpointUrl = if (eventData.startsWith("/")) {
+                Url(baseUrl).protocolWithAuthority + eventData
+            } else {
+                "$baseUrl/$eventData"
+            }
+            endpoint.complete(endpointUrl)
```
kpavlov, line 155: There is a redundant step where the baseUrl is parsed back into Url in the absolute-path branch.

The baseUrl is made up of `url.protocolWithAuthority + path`, where url is `session.call.request.url`. For the absolute-path branch, you only need the origin, so the path part is not needed. By using a string-parse (`Url(baseUrl)`) to get `.protocolWithAuthority`, you are doing extra work that you already have and creating unnecessary memory allocation every time an endpoint event happens.

Instead, you could separate the two properties that the class actually needs:
```kotlin
  private val origin: String by lazy { session.call.request.url.protocolWithAuthority }

  private val baseUrl: String by lazy {
      val url = session.call.request.url
      val path = url.encodedPath
      when {
          path.isEmpty() -> origin
          path.endsWith("/") -> origin + path.removeSuffix("/")
          else -> origin + path.take(path.lastIndexOf("/"))
      }
  }
```

Then handleEndpoint becomes:
```kotlin
val endpointUrl = if (eventData.startsWith("/")) origin + eventData else "$baseUrl/$eventData"
```

kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/MockSseClientEngine.kt
```diff
@@ -0,0 +1,99 @@
...
+            onPostRequest(data)
+            respondWithAccepted()
+        }
+
+        else -> error("Unsupported method: ${data.method}")
+    }
+
+    override fun close() {
+        channel.close()
+        job.complete()
```
kpavlov: `Job.complete()` signals that no more children will be added and transitions to the Completing state; it does not cancel anything. For a `SupervisorJob`, `complete()` will not actually complete until all its children are done. If the test times out before channel's consumers finish, the job remains live and
  the coroutine context leaks across tests.

```suggestion
        job.cancel()
```

kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/SseClientTransportTest.kt
Amaneusz: Nice - added handling and test for full url reolution

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
Amaneusz: Good idea - refactored the code to make it cleaner

kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/MockSseClientEngine.kt
Amaneusz: Good catch, changed

kpavlov (approved):
LGTM

Issue #406, opened 2025-11-19
Align Kotlin SDK with 2025-11-25 Specification


kpavlov:
## Overview

The Kotlin SDK currently implements protocol version `2025-06-18` but is missing several features defined in the `2025-11` TypeScript specification. This issue tracks the discrepancies and required implementations.

<details>
<summary>AI-generated tasks</summary>

### 1. Protocol Version ❌

**Current:** `LATEST_PROTOCOL_VERSION = "2025-06-18"`
**Required:** `LATEST_PROTOCOL_VERSION = "2025-11-25"`

**Files:**
- `kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/common.kt:13`

---

### 2. Task Augmentation System ❌ (Complete Feature Missing)

The entire task system is absent from the Kotlin SDK.

#### 2.1 Missing Core Types

- [ ] `TaskMetadata` - Metadata for task-augmented requests
- [ ] `Task` - Task state data (taskId, status, createdAt, ttl, etc.)
- [ ] `TaskStatus` - Enum: `working | input_required | completed | failed | cancelled`
- [ ] `RelatedTaskMetadata` - For associating messages with tasks

#### 2.2 Missing RequestParams Field

- [ ] Add `task: TaskMetadata?` field to `RequestParams` interface
  - **File:** `kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/request.kt:29`
  - Update all implementing classes

#### 2.3 Missing Request/Response Types

- [ ] `CreateTaskResult` - Response to task-augmented requests
- [ ] `GetTaskRequest` / `GetTaskResult` - Retrieve task state
- [ ] `GetTaskPayloadRequest` / `GetTaskPayloadResult` - Retrieve completed task result
- [ ] `CancelTaskRequest` / `CancelTaskResult` - Cancel a running task
- [ ] `ListTasksRequest` / `ListTasksResult` - List all tasks

#### 2.4 Missing Notifications

- [ ] `TaskStatusNotification` - Notification of task status changes
- [ ] `TaskStatusNotificationParams`

#### 2.5 Missing DSL Builders

- [ ] Update `RequestBuilder` to support task metadata
- [ ] `GetTaskRequestBuilder`
- [ ] `ListTasksRequestBuilder`
- [ ] `CancelTaskRequestBuilder`

---

[...]

PR #566, opened 2026-02-26
feat: add tasks models

### LINKED ISSUE #421: Implement SEP-1686: Tasks
This is a tracking issue for implementation of [SEP-1686](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1686).

## Summary

This SEP introduces the task primitive and task ID to support workflow-based operations in MCP, enabling call-now, fetch-later execution patterns across all request types. Tasks are durable state machines with five states (working, input_required, completed, failed, cancelled) that allow requestors to augment requests with task metadata for deferred result retrieval via polling. This addresses critical use cases for long-running operations such as data analysis pipelines, enterprise automation, code migrations, test execution, deep research, and multi-agent systems where immediate responses are impractical and result retention is needed for client-driven status checking.

The Kotlin SDK currently has no support for tasks or async execution patterns. This implementation will require extensive changes including: new task primitive types (Task, CreateTaskResult, TaskStatus enum), task-related request/response types for tasks/get, tasks/result, tasks/list, and tasks/cancel operations, capability declarations in ClientCapabilities and ServerCapabilities structured by request category (tools.call, sampling.createMessage, elicitation.create), optional task field in existing request parameters to support task-augmented execution, metadata handling for io.modelcontextprotocol/related-task associations, TTL-based resource management, status polling mechanisms, DSL builders for task operations, and comprehensive test coverage. The implementation must be backward compatible, allowing receivers to ignore task metadata if not supported, and providing flexibility for receivers to support both traditional immediate execution and task-augmented patterns.

- Add types:
  - `Task`
  - `TaskStatus`
  - `TaskMetadata`
  - `RelatedTaskMetadata`
  - `CreateTaskResult`
  - `ListTasksRequest`
  - `ListTasksResult`
  - `GetTaskRequest`
  - `GetTaskResult`
  - `GetTaskPayloadRequest`
  - `GetTaskPayloadResult`
  - `CancelTaskRequest`
  - `CancelTaskResult`
  - `TaskStatusNotification`
  - `TaskStatusNotificationParams`
- Add methods to `Method.Defined`.
- Update serializers and DSL builders.

devcrocod:
## Motivation and Context

closes #421 
#406 

## How Has This Been Tested?
Unit tests

## Breaking Changes
NaN

## Types of changes
- [ ] Documentation update

## Checklist


kotlin-sdk-core/detekt-baseline-commonTestSourceSet.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>LargeClass:TasksTest.kt:TasksTest</ID>
```
kpavlov: Let's not inflate detekt baselines.
In this case the test class should be splitted

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/request.kt
```diff
@@ -22,6 +23,17 @@ public value class RequestMeta(public val json: JsonObject) {
...
 
+    /**
+     * The related task metadata, if this request is associated with a task.
+     *
+     * @see RelatedTaskMetadata
+     * @see RELATED_TASK_META_KEY
+     */
+    public val relatedTask: RelatedTaskMetadata?
+        get() = json[RELATED_TASK_META_KEY]?.let { element ->
+            McpJson.decodeFromJsonElement(element)
```
kpavlov, line 34: Could we cache it or handle during deserialization in the serializer? Currently, it will run json parsing every time

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/notification.kt
```diff
@@ -348,3 +348,48 @@ public data class ToolListChangedNotification(override val params: BaseNotificat
...
+ * @property meta Optional metadata for this notification.
+ */
+@Serializable
+public data class TaskStatusNotificationParams(
+    override val taskId: String,
+    override val status: TaskStatus,
+    override val statusMessage: String? = null,
+    override val createdAt: String,
+    override val lastUpdatedAt: String,
+    override val ttl: Double?,
```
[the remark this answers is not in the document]
kpavlov: If other fields have default values here, wouldn't it be nice to define ttl the same way?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tasks.kt
```diff
@@ -0,0 +1,347 @@
...
+    override val lastUpdatedAt: String,
+    override val ttl: Double?,
+    override val pollInterval: Double? = null,
+    @SerialName("_meta")
+    override val meta: JsonObject? = null,
+) : ClientResult,
+    ServerResult,
+    TaskFields
+
+// ============================================================================
```
kpavlov, line 181: This file is getting huge. Let's split it. There are clear boundaries: models, requests, responses

kotlin-sdk-core/detekt-baseline-main.xml
```diff
@@ -3,6 +3,7 @@
   <ManuallySuppressedIssues/>
   <CurrentIssues>
     <ID>CyclomaticComplexMethod:jsonUtils.kt:@OptIn(ExperimentalUnsignedTypes::class, ExperimentalSerializationApi::class) private fun convertToJsonElement: JsonElement</ID>
+    <ID>CyclomaticComplexMethod:serializers.kt:private fun selectServerResultDeserializer: DeserializationStrategy&lt;ServerResult>?</ID>
```
kpavlov: Can we address it in a proper way?

kpavlov (changes requested):
Thank you, @devcrocod 
There are copilot findings that look legitimate.
Let's split big classes and don't inflate detekt baselines.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
```diff
@@ -301,6 +311,9 @@ private fun selectClientResultDeserializer(element: JsonElement): Deserializatio
         "model" in jsonObject && "role" in jsonObject -> CreateMessageResult.serializer()
         "roots" in jsonObject -> ListRootsResult.serializer()
         "action" in jsonObject -> ElicitResult.serializer()
+        "task" in jsonObject -> CreateTaskResult.serializer()
+        "tasks" in jsonObject -> ListTasksResult.serializer()
+        "taskId" in jsonObject -> GetTaskResult.serializer()
```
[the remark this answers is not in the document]
Amaneusz: Additionally, I think we should include tests for polymorphic deserialization that would have to trigger logic  `selectClientResultDeserializer` in `serializers.kt`.
Currently deserialization tests from `TasksTest` such as `should deserialize CancelTaskResult and verify TaskFields contract` use specific deserializers:
```
// ...
val result = verifyDeserialization<CancelTaskResult>(McpJson, json)
// ...
```

This only tests if deserializers work correctly but not if the correct deserializer was chosen during the selection phase.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/notification.kt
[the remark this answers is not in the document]
devcrocod: ttl isn't an optional property
https://modelcontextprotocol.io/specification/2025-11-25/schema#task

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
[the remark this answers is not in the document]
devcrocod: Thank you for pointing that out 
There is validation in place for the other polymorphic serializers.
I’ll review the field-to-type mappings again and add tests to cover these cases

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/request.kt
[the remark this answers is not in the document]
devcrocod: In my opinion, we should throw an error here. Returning null if relatedTask does not match RelatedTaskMetadata is not the best option

devcrocod: json is not parsed every time, It’s not an expensive operation

I can do it that way, but then we’ll have to modify the class since it’s `value class`

devcrocod: > handle during deserialization in the serializer

And that would require a custom serializer, which I’d prefer to avoid

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tasks.kt
```diff
@@ -0,0 +1,347 @@
...
+ * @property meta Optional metadata for this response.
+ */
+@Serializable
+public data class GetTaskResult(
+    override val taskId: String,
+    override val status: TaskStatus,
+    override val statusMessage: String? = null,
+    override val createdAt: String,
+    override val lastUpdatedAt: String,
+    override val ttl: Double?,
```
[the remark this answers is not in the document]
devcrocod: ttl is not optional 
https://modelcontextprotocol.io/specification/2025-11-25/schema#task

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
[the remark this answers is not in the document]
devcrocod: I looked into this more closely.
[CancelTaskResult](https://modelcontextprotocol.io/specification/2025-11-25/schema#canceltaskresult) and [GetTaskResult](https://modelcontextprotocol.io/specification/2025-11-25/schema#gettaskresult) have the same schema
That means their json will be identical. We determine the serializer based on the json rpc method, so a polymorphic serializer should not need to be used

kotlin-sdk-core/detekt-baseline-commonTestSourceSet.xml
devcrocod: https://github.com/modelcontextprotocol/kotlin-sdk/pull/567#discussion_r2871591249

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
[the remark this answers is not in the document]
devcrocod: I made CancelTaskResult as a type alias of GetTaskResult.
added a test

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/TasksTest.kt
```diff
@@ -443,13 +446,33 @@ class TasksTest {
...
+              ],
+              "isError": false
+            }
+        """.trimIndent()
+
+        val result = verifyDeserialization<GetTaskPayloadResult>(McpJson, json)
+        assertNotNull(result.meta)
+        assertEquals("task-42", result.meta?.get("origin")?.jsonPrimitive?.content)
+        assertNotNull(result["content"])
+        assertNotNull(result["isError"])
```
kpavlov: Why not give Kotest a try for our new tests?

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/TasksTest.kt
```diff
@@ -742,4 +765,27 @@ class TasksTest {
...
+        fun <T : RequestResult> uncheckedCast(value: RequestResult): T {
+            @Suppress("UNCHECKED_CAST")
+            return value as T
+        }
+
+        val cancelResult: CancelTaskResult = uncheckedCast(result)
+
+        assertEquals("task-1", cancelResult.taskId)
+        assertEquals(TaskStatus.Cancelled, cancelResult.status)
+    }
```
kpavlov, line 801: Wow, that’s almost 800 lines!

How about we break things down a bit? We could split tasks.kt and TaskTest.kt into different files, like GetTask.kt, ListTasks.kt and CancelTask.kt, based on what each one does.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tasks.kt
```diff
@@ -228,11 +230,22 @@ public data class GetTaskPayloadRequestParams(
  *
  * @property meta Optional metadata for this response.
  */
+@JvmInline
 @Serializable
-public data class GetTaskPayloadResult(
-    @SerialName("_meta")
-    override val meta: JsonObject? = null,
-) : ServerResult
+public value class GetTaskPayloadResult(public val json: JsonObject) :
```
kpavlov, line 235: I am not sure I understand the advantages of moving to the value class here. It seems like it might not be possible to extend it by the user.
@e5l, do you think we are safe here?

kpavlov:
@devcrocod, please consider defining a `public typealias TaskId = String` or value class for type safety and, potentially, validation

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/request.kt
kpavlov: > 
json is not parsed every time, It’s not an expensive operation

I see the getter is not cached, so every property call will result in json parsing. In general, it should be avoided. The application would have to cache it. 

A value class may not be the best solution here, as it introduces unnecessary restrictions.

Additionally, `RequestMeta` risks becoming a _God Object_ as more metadata (e.g., `RelatedTaskMetadata`) is added over time. This pattern hinders extensibility and violates clean design principles.

An alternative solution is to use an extension function:

```kotlin
public fun RequestMeta.relatedTaskMetadata(): RelatedTaskMetadata? = this.json[RELATED_TASK_META_KEY]?.let {
    McpJson.decodeFromJsonElement(it)
}
```

This approach is more extensible and avoids the pitfalls of the current solution, even if it doesn't address the lack of caching.

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/TasksTest.kt
```diff
@@ -0,0 +1,791 @@
...
+            {
+              "taskId": "task-42"
+            }
+            """.trimIndent(),
+        )
+    }
+
+    @Test
+    fun `RELATED_TASK_META_KEY should have correct value`() {
+        assertEquals("io.modelcontextprotocol/related-task", RELATED_TASK_META_KEY)
```
kpavlov: This test is not valuable. Let's verify RequestMeta deserialization with RelatedTaskMetadata instead.

Would it be possible to get some test data from the spec? https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks, e.g. https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#retrieving-task-results response is pretty good.

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/TasksTest.kt
```diff
@@ -0,0 +1,791 @@
...
+        )
+    }
+
+    @Test
+    fun `should deserialize GetTaskRequest`() {
+        val json = """
+            {
+              "method": "tasks/get",
+              "params": {
+                "taskId": "task-11"
```
kpavlov, line 323: Are we missing the "io.modelcontextprotocol/related-task" here?

> [5.7.1.​](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#associating-task-related-messages)
Associating Task-Related Messages
All requests, notifications, and responses related to a task MUST include the io.modelcontextprotocol/related-task key in their _meta field, with the value set to an object with a taskId matching the associated task ID.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/request.kt
devcrocod: > I see the getter is not cached, so every property call will result in json parsing. In general, it should be avoided. The application would have to cache it.

To clarify, `decodeFromJsonElement` doesn't parse a json string. The `JsonObject` tree is already built at that point, so this call simply maps an existing JsonElement to a `RelatedTaskMetadata(taskId: String)`, essentially one string read from an in-memory tree. It's a very lightweight operation.

> I see the getter is not cached, so every property call will result in json parsing. In general, it should be avoided. The application would have to cache it.

`RequestMeta` currently has only two properties: `progressToken` and `relatedTask`. Both are defined by the MCP spec, not arbitrary extensions. I don't think two spec-defined accessors qualify as a God Object pattern

The relatedTask property was intentionally implemented to be consistent with `progressToken`, which is already defined as a computed property on `RequestMeta` in the same style. Moving one to an extension function while keeping the other as a property would break that consistency

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/TasksTest.kt
devcrocod: No, this is correct:

> For the tasks/get, tasks/result, and tasks/cancel operations, the taskId parameter in the request MUST be used as the source of truth for identifying the target task. Requestors SHOULD NOT include io.modelcontextprotocol/related-task metadata in these requests, and receivers MUST ignore such metadata if present in favor of the RPC method parameter. Similarly, for the tasks/get, tasks/list, and tasks/cancel operations, receivers SHOULD NOT include io.modelcontextprotocol/related-task metadata in the result messages, as the taskId is already present in the response structure.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
```diff
@@ -301,6 +311,9 @@ private fun selectClientResultDeserializer(element: JsonElement): Deserializatio
         "model" in jsonObject && "role" in jsonObject -> CreateMessageResult.serializer()
         "roots" in jsonObject -> ListRootsResult.serializer()
         "action" in jsonObject -> ElicitResult.serializer()
+        "task" in jsonObject -> CreateTaskResult.serializer()
```
kpavlov, line 314: Should we need to add GetTaskPayloadResult somewhere?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tasks.kt
```diff
@@ -0,0 +1,347 @@
...
+/**
+ * Common fields shared by all types representing the task state.
+ */
+public sealed interface TaskFields {
+    public val taskId: String
+    public val status: TaskStatus
+    public val statusMessage: String?
+    public val createdAt: String
+    public val lastUpdatedAt: String
+    public val ttl: Double?
```
kpavlov: > ttl: number | null
> Actual retention duration from creation in milliseconds, null for unlimited.

Let's make it `Long`, not `Double`

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/NotificationTest.kt
```diff
@@ -310,4 +310,116 @@ class NotificationTest {
...
+                "pollInterval": 1000.0
+              }
+            }
+        """.trimIndent()
+
+        val notification = verifyDeserialization<TaskStatusNotification>(McpJson, json)
+        val params = notification.params!!
+
+        assertEquals(Method.Defined.NotificationsTasksStatus, notification.method)
+        assertEquals("task-3", params.taskId)
```
kpavlov, line 402: nit: please, consider switching to kotest assertions

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tasks.kt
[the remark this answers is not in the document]
kpavlov: ttl is nullable (representing "no limit") but requires the caller to write ttl = null at every call site because it has no default. This asymmetry with pollInterval (same pattern, has default) is inconsistent, adds noise at call sites (the tests show ttl = null repeated ~15 times)

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
devcrocod: Yes, but that would require refactoring protocol, so it's out of scope for this pr

kotlin-sdk-core/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/types/NotificationTest.kt
devcrocod: In NotificationTest, kotlin.test is used, so I keep it for consistency
We can update it across the entire file later

PR #567, opened 2026-02-26
feat: add tool execution properties to Tool

### LINKED ISSUE #543: Add ToolExecution to Tool definition
[ToolExecution](https://modelcontextprotocol.io/specification/2025-11-25/schema#toolexecution)

Tools can declare task-augmented execution support via `ToolExecution`

devcrocod:
closes #543 

## How Has This Been Tested?
unit

## Breaking Changes
binary compatibility in the Tool class

## Types of changes
- [ ] Documentation update

## Checklist


kotlin-sdk-core/detekt-baseline-test.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<SmellBaseline>
+  <ManuallySuppressedIssues/>
+  <CurrentIssues>
+    <ID>LargeClass:ToolsTest.kt:ToolsTest</ID>
```
kpavlov: 1. This is a legitimate finding. Let's split ToolsTest for maintainability instead of suppressing warning.

2. In general, let's not add anything to baselines. Only remove. In exceptional cases let's use `@Suppress annotation in place`

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
```diff
@@ -1618,7 +1618,7 @@ public fun Tool(
     title,
     annotations,
     icons,
-    _meta,
+    meta = _meta,
```
kpavlov, line 1621: Why not to add other parameter names for consistency?

kotlin-sdk-core/detekt-baseline-test.xml
devcrocod: 1. currently, our approach is to group unit tests by entity for types. Splitting ToolsTest into separate files would increase complexity and make the structure less clear 
2. ideally, we should avoid adding `@Suppress` annotations from third-party tools into the source code. For now, I've added this to baseline, which is also not the best solution. If you see a better approach, I'd appreciate your suggestion

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
devcrocod: I added this to avoid a compilation error. There’s not much point in adding it for the others, since they are already annotated as @Deprecated with an error level

kotlin-sdk-core/detekt-baseline-test.xml
kpavlov: 1. This method might not be the best fit when the group gets too large.

> Splitting ToolsTest into separate files would increase complexity and make the structure less clear

Actually, it is often the other way around. Smaller files are usually easier to understand and keep up with. This one is over 500 lines and checks out several classes. It could be broken down into smaller pieces, so let’s do that.

kpavlov: 2. Ideally, we should avoid situations where `@Suppress` is needed.

There is one exception: when we are clearly explaining our design choice to NOT follow best practices. We can add a comment to clarify this. If it is in the code, it is easy to see whether it is a false positive or not. But if it is in the baseline file, it’s separate from the code.

Let’s focus on removing baseline files rather than adding more to them.

I’m afraid I won’t be able to give +1 on this.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types.kt
kpavlov: I find consistency and readability pretty good reasons, but they don’t stop me from moving forward. Instead, it just give me a little whiff.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/tools.kt
```diff
@@ -92,6 +95,38 @@ public data class ToolSchema(
...
 
+/**
+ * Execution-related properties for a tool.
+ *
+ * @property taskSupport Indicates whether this tool supports task-augmented execution.
+ * This allows clients to handle long-running operations through polling the task system.
+ * Defaults to [TaskSupport.Forbidden] when absent.
+ */
+@Serializable
+public data class ToolExecution(val taskSupport: TaskSupport? = null)
```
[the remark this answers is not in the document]
devcrocod: optional in mcp schema: https://modelcontextprotocol.io/specification/2025-11-25/schema#toolexecution

Issue #564, opened 2026-02-23
Stdio server transport is in a bad state


i1bro:
**Describe the bug**
Stdio transport is the most convenient option in a distributed application, since you can pass a custom sink/source to it. However, the current implementation has a bunch of critical bugs:

1) Client messages are processed sequentially. This means there can't be 2 requests being processed at once.
If there's a pinger to check client liveness, the "pong" response from client won't be handled before the previous, maybe very long-running, request handler is finished.
Also it blocks advanced protocol features like elicitation.

2) Request handlers are launched on Dispatchers.IO

3) The coroutine scope that's used in this transport is completely detached from the coroutine tree, which may cause leaks. Please add an optional CoroutineScope parameter

4) Cancellation exceptions are being caught and not rethrown

rnett:
@kpavlov @devcrocod @e5l https://github.com/modelcontextprotocol/kotlin-sdk/pull/670 was closed without comment and I don't see any similar PRs.  Is this still being worked on?  In the current state of this library STDIO is rather broken (w/ this and https://github.com/modelcontextprotocol/kotlin-sdk/issues/572), and it's still probably the most common MCP transport scheme.

devcrocod:
@rnett 
this issue will be addressed in separate PRs
Both this issue and the other one you mentioned are high priority for us

Issue #242, opened 2025-08-25
Cancellation exceptions cause an error to be logged


phemmer:
**Describe the bug**
When the coroutine running a handler is cancelled, it throws a `kotlinx.coroutines.JobCancellationException`. The MCP SDK is picking this up and logging this as an error, instead of gracefully ignoring it. To exacerbate the issue, in the IntelliJ IDE, logs with `error` severity are presented to the user in a notification bubble. This should not be happening for something as mundane as a coroutine being cancelled.

**To Reproduce**
1. Create a MCP server with a handler and have that handler throw a `java.util.concurrent.CancellationException` or any subclass of.

**Expected behavior**
The coroutine should gracefully shut down without an error being logged.

**Logs**
```
Error handling request: tools/list (id: NumberId(value=1))
kotlinx.coroutines.JobCancellationException: Job was cancelled; job=SupervisorJobImpl{Cancelled}@695a8e2e
```

And yes, that is the extent of the stack trace, as cancellation exceptions suppress the stack trace.

**Additional context**
In one specific example I'm looking at, the issue is happening [here](https://github.com/modelcontextprotocol/kotlin-sdk/blob/b37a0f62d03da80e1261be165ef6992df705b06a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt#L283). However the code has this issue all over the place.

----

While this issue is about `CancellationException` being logged, I might argue that a different approach for logging is needed. As mentioned, IntelliJ IDEs present any logs with error level to the user in a notification bubble. Other apps may treat logs differently. Thus the application using the SDK needs to be able to control them. I might argue that the SDK shouldn't emit logs (at least errors) at all, and that it should be raising exceptions which the application can decide what to do with. Another option might be to allow overriding the logger, without having to resort to all sorts of shenanigans to hijack the SDK's call to `KotlinLogging.logger`, or hook into slf4j, etc.

kpavlov:
The logging defines how the support team should react:

- **ERROR** – Action should be taken immediately! Ops team should enable [Rollbar](https://rollbar.com/) + [PagerDuty](https://www.pagerduty.com/) or other notification service in order to receive and alert when an ERROR appears in logs.
- **WARN** – Certain action should be taken, but it can wait till the next business day.
- **INFO** (enabled by default) — Use this to print information you want to see in logs when your application works normally.
[...]

PR #571, opened 2026-03-02
fix!(server): #575 Rethrow CancellationException in StdioServerTransport, fix ReadBuffer

kpavlov:
### Changes
1. StdioServerTransport
    - Replace generic `Throwable` catches with specific `CancellationException` and `Exception` for clarity and correctness. **breaking change** `Throwable` is not handled any more!
    - Refactor launching jobs to separate methods
    - Introduce `READ_BUFFER_SIZE` constant to replace inline buffer size literals.
    - Add suppress annotations for clearer intent and constructor documentation for better usability.
    - Extend StdioServerTransportTest

2. ReadBuffer

    Previously, readMessage() returned null after consuming an unparseable line even when more complete lines existed in the buffer. This caused valid messages following a bad line in the same chunk to be silently dropped until the next chunk arrived (or forever, in tests).
    
    Fix the loop so null means only "no complete line available", not "encountered a parse failure". Blank/whitespace-only lines are now silently skipped via isBlank() rather than forwarded to the deserializer and logged as errors.
    
    Refactor the method into three focused helpers:
     - readMessage() — outer loop over lines
     - readLine()    — consume the next newline-delimited line from the buffer
     - tryRecover()  — attempt deserialization from the first '{' onward

3. Add utility method `runIntegrationTest` for integration testing (`runBlocking` + `withTimeout`). Use it in StdioServerTransportTest

4. Store processing Job reference and `calcelAndJoin()` it on close. Partially addressing #574

## Motivation and Context

See #575, #564, #242 

## How Has This Been Tested?
Unit test added

## Breaking Changes

Semantic change: `Throwable` and CancellationException are not handled any more!

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
@@ -87,7 +87,7 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
                 }
             } catch (e: CancellationException) {
                 throw e
-            } catch (e: Exception) {
+            } catch (e: Throwable) {
                 logger.error(e) { "Error reading from stdin" }
                 _onError.invoke(e)
             } finally {
@@ -111,7 +111,7 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
                 }
             } catch (e: CancellationException) {
                 throw e
-            } catch (e: Exception) {
+            } catch (e: Throwable) {
                 _onError.invoke(e)
             }
         }
@@ -132,7 +132,7 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
                 }
             } catch (e: CancellationException) {
                 throw e
-            } catch (e: Exception) {
+            } catch (e: Throwable) {
                 logger.error(e) { "Error writing to stdout" }
                 _onError.invoke(e)
             }
@@ -151,7 +151,7 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
         while (true) {
             val message = try {
                 readBuffer.readMessage()
-            } catch (e: Exception) {
+            } catch (e: Throwable) {
                 _onError.invoke(e)
                 null
             }
@@ -162,7 +162,8 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
                 _onMessage.invoke(message)
             } catch (e: CancellationException) {
                 throw e
-            } catch (e: Exception) {
+            } catch (e: Throwable) {
+                logger.error(e) { "Error processing message" }
                 _onError.invoke(e)
             }
         }
--- a/kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
+++ b/kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
@@ -160,23 +160,23 @@ class StdioServerTransportTest {
     // region: Exception handling
 
     @ParameterizedTest(name = "[{index}] input throws {0}")
-    @MethodSource("inputExceptions")
-    fun `should invoke onError when input stream throws`(exception: Exception): Unit = runIntegrationTest {
-        val server = StdioServerTransport(FaultyRawSource(exception).buffered(), printOutput)
+    @MethodSource("inputErrors")
+    fun `should invoke onError when input stream throws`(throwable: Throwable): Unit = runIntegrationTest {
+        val server = StdioServerTransport(FaultyRawSource(throwable).buffered(), printOutput)
         val capturedError = CompletableDeferred<Throwable>()
         server.onError { capturedError.complete(it) }
         server.onMessage {}
 
         server.start()
 
-        capturedError.await() shouldBe exception
+        capturedError.await() shouldBe throwable
         server.close()
     }
 
     @ParameterizedTest(name = "[{index}] output throws {0}")
-    @MethodSource("outputExceptions")
-    fun `should invoke onError when output sink throws`(exception: Exception): Unit = runIntegrationTest {
-        val server = StdioServerTransport(bufferedInput, FaultyRawSink(exception).buffered())
+    @MethodSource("outputErrors")
+    fun `should invoke onError when output sink throws`(throwable: Throwable): Unit = runIntegrationTest {
+        val server = StdioServerTransport(bufferedInput, FaultyRawSink(throwable).buffered())
         val capturedError = CompletableDeferred<Throwable>()
         server.onError { capturedError.complete(it) }
         server.onMessage {}
@@ -184,7 +184,7 @@ class StdioServerTransportTest {
         server.start()
         server.send(PingRequest().toJSON())
 
-        capturedError.await() shouldBe exception
+        capturedError.await() shouldBe throwable
         server.close()
     }
 
@@ -218,8 +218,8 @@ class StdioServerTransportTest {
     }
 
     @ParameterizedTest(name = "[{index}] handler throws {0}")
-    @MethodSource("handlerExceptions")
-    fun `should continue processing messages after handler throws`(exception: Exception) = runIntegrationTest {
+    @MethodSource("handlerErrors")
+    fun `should continue processing messages after handler throws`(throwable: Throwable) = runIntegrationTest {
         val server = StdioServerTransport(bufferedInput, printOutput)
         val capturedErrors = mutableListOf<Throwable>()
         val receivedMessages = mutableListOf<JSONRPCMessage>()
@@ -231,7 +231,7 @@ class StdioServerTransportTest {
         server.onError { capturedErrors.add(it) }
         server.onMessage { message ->
             if (message == message1) {
-                throw exception
+                throw throwable
             } else {
                 receivedMessages.add(message)
                 secondMessageProcessed.complete(Unit)
@@ -246,7 +246,7 @@ class StdioServerTransportTest {
 
         secondMessageProcessed.await()
 
-        capturedErrors shouldContain exception
+        capturedErrors shouldContain throwable
         receivedMessages shouldBe listOf(message2)
         server.close()
     }
@@ -300,34 +300,37 @@ class StdioServerTransportTest {
     }
 
     @Suppress("unused")
-    private fun inputExceptions() = listOf(
+    private fun inputErrors() = listOf(
         IOException("simulated read failure"),
-        RuntimeException("unexpected read error"),
+        RuntimeException("unexpected read exception"),
+        OutOfMemoryError("unexpected read error"),
     )
 
     @Suppress("unused")
-    private fun outputExceptions() = listOf(
+    private fun outputErrors() = listOf(
         IOException("simulated write failure"),
-        RuntimeException("unexpected write error"),
+        RuntimeException("unexpected write exception"),
+        OutOfMemoryError("unexpected write error"),
     )
 
     @Suppress("unused")
-    private fun handlerExceptions() = listOf(
+    private fun handlerErrors() = listOf(
         RuntimeException("handler failure"),
         IOException("handler IO failure"),
+        OutOfMemoryError("handler error"),
     )
 
-    /** A [RawSource] that immediately throws [exception] on every read attempt. */
-    private class FaultyRawSource(private val exception: Exception) : RawSource {
-        override fun readAtMostTo(sink: Buffer, byteCount: Long): Long = throw exception
+    /** A [RawSource] that immediately throws [throwable] on every read attempt. */
+    private class FaultyRawSource(private val throwable: Throwable) : RawSource {
+        override fun readAtMostTo(sink: Buffer, byteCount: Long): Long = throw throwable
         override fun close() {
             // noop
         }
     }
 
-    /** A [RawSink] that throws [exception] on every [write] call. */
-    private class FaultyRawSink(private val exception: Exception) : RawSink {
-        override fun write(source: Buffer, byteCount: Long): Unit = throw exception
+    /** A [RawSink] that throws [throwable] on every [write] call. */
+    private class FaultyRawSink(private val throwable: Throwable) : RawSink {
+        override fun write(source: Buffer, byteCount: Long): Unit = throw throwable
         override fun flush() {
             // noop
         }


kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -46,17 +55,29 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
     private val writeChannel = Channel<JSONRPCMessage>(Channel.UNLIMITED)
     private val outputWriter = outputStream.buffered()
 
+    @Suppress("TooGenericExceptionCaught", "ThrowsCount")
```
devcrocod: Is this suppress really needed here?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -46,17 +55,29 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
...
+        readingJob = launchReadingJob()
+
+        // Launch a coroutine to process messages from readChannel
+        launchProcessingJob()
+
+        // Launch a coroutine to handle message sending
+        sendingJob = launchSendingJob()
+    }
+
+    @Suppress("TooGenericExceptionCaught")
```
devcrocod: These annotations just add noise to the code, especially considering they’re by external tooling

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -66,47 +87,71 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
...
     }
 
+    @Suppress("TooGenericExceptionCaught")
     private suspend fun processReadBuffer() {
         while (true) {
             val message = try {
                 readBuffer.readMessage()
-            } catch (e: Throwable) {
+            } catch (e: CancellationException) {
+                throw e
```
devcrocod: Is there actually any scenario where `CancellationException` could be thrown here?

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
```diff
@@ -135,7 +156,153 @@ class StdioServerTransportTest {
...
+    }
+
+    @Test
+    fun `should throw when starting twice`(): Unit = runBlocking {
+        withTimeout(TEST_TIMEOUT) {
+            val server = StdioServerTransport(bufferedInput, printOutput)
+            server.onMessage {}
+            server.start()
+            try {
+                assertFailsWith<IllegalStateException> {
```
devcrocod: If we’re using kotest, I think we can also use `shouldThrow`

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
```diff
@@ -135,7 +156,153 @@ class StdioServerTransportTest {
...
+
+            server.start()
+
+            received.await() shouldBe validMessage
+            server.close()
+        }
+    }
+
+    companion object {
+        @JvmStatic
```
devcrocod: Why do we need these static?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
```diff
@@ -20,45 +20,69 @@ public class ReadBuffer {
...
+     *
+     * @return A deserialized `JSONRPCMessage` if successfully processed; otherwise, `null` if the
+     *         input buffer is exhausted or no valid message is found.
+     */
     public fun readMessage(): JSONRPCMessage? {
-        if (buffer.exhausted()) return null
-        var lfIndex = buffer.indexOf('\n'.code.toByte())
-        val line = when (lfIndex) {
-            -1L -> return null
+        var message: JSONRPCMessage? = null
```
devcrocod: I think we don’t need to use a mutable property here
btw that’s not an issue

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -30,10 +31,18 @@ import kotlin.coroutines.CoroutineContext
...
+ *
+ * @constructor Creates a new instance of [StdioServerTransport].
+ * @param inputStream The input stream used to receive data.
+ * @param outputStream The output stream used to send data.
  */
 @OptIn(ExperimentalAtomicApi::class)
 public class StdioServerTransport(private val inputStream: Source, outputStream: Sink) : AbstractTransport() {
-    private val logger = KotlinLogging.logger {}
+
+    private companion object {
```
devcrocod: In my opinion, creating a separate companion object here is unnecessary. It’s only needed if we want to access those fields from outside the class, not just within it

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: Idea/detekt complains about throwing `Exception`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: It might be when ReadBuffer will be cancelable. Just to be safe in the future

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: JUnit wants `@MethodSource` metods to be static

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: true, let me change this in the next PR

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: There is no catch exception here.
Only in other functions that already have `@Suppress`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: function is not suspend

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
devcrocod: Isn’t a companion object enough?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: right, I'll remove it

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: I'll move them next to the `try` block. I'd say it's not a noise but a reminder why we've made a concious decision to catch all exceptions

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: removed companion object

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: it should be static mehtod arrording to junit contract. A [standard pattern](https://stackoverflow.com/a/76738406) for junit parametrized tests.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
kpavlov: fixed

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: fixed

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: done

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: fixed

kpavlov:
Thank you, @devcrocod, for your comments! I tried to address most of them, please have a look

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: Used `@TestInstance(TestInstance.Lifecycle.PER_CLASS)` and migrated from static to instance methods

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -133,13 +195,15 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
             }.onFailure { logger.warn(it) { "Failed to close stdin" } }
 
             readingJob?.cancel()
-
             readChannel.close()
+
+            processingJob?.cancelAndJoin()
```
kpavlov, line 201: #574

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ReadBuffer.kt
```diff
@@ -20,45 +20,70 @@ public class ReadBuffer {
...
+            val line = readNextLine() ?: return null
+            if (line.isBlank()) continue
 
-            0L -> {
-                buffer.skip(1)
-                return null
+            @Suppress("TooGenericExceptionCaught")
+            val message = try {
+                deserializeMessage(line)
+            } catch (e: Exception) {
```
e5l, line 44:
```suggestion
            } catch (e: Throwable) {
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -66,47 +85,73 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
...
                     val json = serializeMessage(message)
-                    outputWriter.writeString(json)
-                    outputWriter.flush()
+                    outputSink.writeString(json)
+                    outputSink.flush()
                 }
-            } catch (e: Throwable) {
+            } catch (e: CancellationException) {
+                throw e
+            } catch (e: Exception) {
```
e5l:
```suggestion
            } catch (e: Throwable) {
```

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -66,47 +85,73 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
...
+        return job
     }
 
     private suspend fun processReadBuffer() {
+        @Suppress("TooGenericExceptionCaught")
         while (true) {
             val message = try {
                 readBuffer.readMessage()
-            } catch (e: Throwable) {
+            } catch (e: Exception) {
```
e5l: Could you tell me why we want to skip logging `Error`? https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html

kpavlov: Things like OutOfMemoryError are not actionable and usually not handled by the application and propagated.

e5l: they still needs to be logged by the configured logger

kpavlov: Logging is never guaranteed in the situation when JVM is crashing.
E.g., an `OutOfMemoryError` usually means the JVM has insufficient memory to allocate additional objects. Application loggers often allocate new objects (e.g., strings, log event objects, buffers) or interact with subsystems (e.g., disk I/O, network I/O). These operations may fail or behave unpredictably under memory pressure. 

Logging to System.err might make sense, but is still questionable.

kpavlov:
@e5l, I’ve reverted to using Throwables. Could you take a look?

PR #579, opened 2026-03-02
ci: add Codecov coverage upload and XML report generation

kpavlov:
## Add Codecov coverage upload and XML report generation

- Add `koverXmlReport` to Gradle tasks in Ubuntu workflow
- Configure Codecov action to upload coverage reports

Reports will be available on https://app.codecov.io/gh/modelcontextprotocol/kotlin-sdk

## Motivation and Context
To provide better test coverage visibility

## How Has This Been Tested?
CI

## Breaking Changes
No

## Types of changes
- [x] CI update

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         include:
           - os: ubuntu-latest
             job-name: "Linux Tests"
-            gradle-tasks: "build ktlintCheck koverLog koverHtmlReport publishToMavenLocal -Pversion=1-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


devcrocod:
Why do we need this?

We shouldn’t introduce a third-party AI tool without a clear and explicit need!

Especially considering that, instead of codecov, we could add the claude plugin or qodana, which would provide similar functionality. And likely in a more reliable way than an external AI tool. And that’s only if we actually need this in the first place

kpavlov:
@devcrocod, just so you know, this isn’t an AI tool. It’s simply a viewer for the generated coverage XML report that Kover creates.

I don’t think there’s anything wrong with adding a reporting tool. What I’m more worried about is the lack of reporting and visibility.

Qodana is a bit slow, and I’m happy to consider switching to another tool that offers better UX.

Issue #574, opened 2026-03-02
StdioServerTransport: The coroutine scope that's used in this transport is completely detached from the coroutine tree


kpavlov:
**Describe the bug**

The coroutine scope that's used in this transport is completely detached from the coroutine tree, which may cause leaks. Please add an optional CoroutineScope parameter

**Additional context**
#564

Issue #573, opened 2026-03-02
StdioTransport:  Request handlers should not run on Dispatchers.IO


kpavlov:
**Describe the bug**
Request handlers are launched on Dispatchers.IO

**Expected behavior**
Message handling should run on default dispatcher by default with an option for the user to provide a custom dispatcher.

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**
#564

Issue #518, opened 2026-02-12
refactor: Unify workflow across server transports


kpavlov:
refactor: Unify workflow across server transports

Currently, there is duplication of workflow-management logic in server transports: atomic booleans for initialized/closed state, session management, etc

Similar to #517, but for server transports. 

- Consider unifying the client/server TransportState enum.
- Consider moving more logic to `AbstractTransport`

## Todo
- Introduce AbstractServerTransport for handling common workflow logic, similar to AbstractClientTransport
- WebSocket transport hierarchy should be refactored
- Introduce transport state + state machine (FSM) for possible transitions
- [TBD]
- Profit

## Definition of done

- Current integration tests pass. Exception types might be changed for unification.
- Unit test for edge cases, especially for the state machine

See #517

PR #584, opened 2026-03-03
fix(stdio): #573, #574 customize dispatchers & scope for stdio server transport configuration

kpavlov:
1. Add a customizable builder to StdioServerTransport

    Scope lifecycle is fixed (#574 resolved), processing dispatcher defaults to `Dispatchers.Default` (#573 resolved), scope context is clean (no spurious dispatcher stacking), and `@Volatile` on the three job vars addresses the visibility race.

    - Introduce a `Configuration` class for `StdioServerTransport` to improve API flexibility and readability. Now one can provide a dispatcher for IO running on Virtual Threads
    - Updated transport initialization to use a builder block for configuring parameters such as I/O streams, buffer sizes, dispatchers, and parent coroutine scope.

    - Added integration test validating the builder functionality.

2. Added `InjectDispatcher` rule to detekt exclusions for test folders in `detekt.yml`. Removed `InjectDispatcher` issues from detekt baseline files.

**NB!** This PR doesn't fully resolve race conditions on close(). It will be resolved in the course of #518

## Motivation and Context
#573, #574

## How Has This Been Tested?

Integration test / regression tests

## Breaking Changes
No, only API is only extended

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
@@ -14,7 +14,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancelAndJoin
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
@@ -25,7 +25,6 @@ import kotlinx.io.Source
 import kotlinx.io.buffered
 import kotlinx.io.readByteArray
 import kotlinx.io.writeString
-import kotlin.concurrent.Volatile
 import kotlin.concurrent.atomics.AtomicBoolean
 import kotlin.concurrent.atomics.ExperimentalAtomicApi
 
@@ -34,101 +33,66 @@ private const val READ_BUFFER_SIZE = 8192L
 /**
  * A server transport that communicates with a client via standard I/O.
  *
- * Reads from input [Source] and writes to output [Sink].
+ * [StdioServerTransport] manages the communication between a JSON-RPC server and its clients
+ * by reading incoming messages from the specified [Source] (input stream) and writing outgoing
+ * messages to the [Sink] (output stream).
  *
  * Example:
  * ```kotlin
- * val transport = StdioServerTransport {
- *   source = System.`in`.asInput(),
- *   sink =  System.out.asSink(),
- * }
+ * val transport = StdioServerTransport(
+ *   source = System.`in`.asInput()
+ *   sink =  System.out.asSink()
+ * )
  * ```
  *
- * @constructor Initializes the transport using the provided block for [Configuration].
- * The configuration includes specifying the input and output streams, buffer sizes,
- * and dispatchers for I/O and processing tasks and coroutine scope.
+ * @constructor Creates an instance of [StdioServerTransport] with the specified parameters.
+ * @property source The source for reading incoming messages (e.g., stdin or other readable stream).
+ * @param sink The sink for writing outgoing messages (e.g., stdout or other writable stream).
+ * @property readBufferSize The maximum size of the read buffer, defaults to a pre-configured constant.
+ * @property readChannel The channel for receiving raw byte arrays from the input stream.
+ * @property writeChannel The channel for sending serialized JSON-RPC messages to the output stream.
+ * @property readingJobDispatcher The dispatcher to use for the message-reading coroutine.
+ * @property writingJobDispatcher The dispatcher to use for the message-writing coroutine.
+ * @property processingJobDispatcher The dispatcher to handle processing of read messages.
+ * @param coroutineScope Optional coroutine scope to use for managing internal jobs. A new scope
+ *              will be created if not provided.
  */
 @OptIn(ExperimentalAtomicApi::class)
-public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTransport() {
+@Suppress("LongParameterList")
+public class StdioServerTransport(
+    private val source: Source,
+    sink: Sink,
+    private val readBufferSize: Long = READ_BUFFER_SIZE,
+    private val readChannel: Channel<ByteArray> = Channel(Channel.UNLIMITED),
+    private val writeChannel: Channel<JSONRPCMessage> = Channel(Channel.UNLIMITED),
+    private var readingJobDispatcher: CoroutineDispatcher = IODispatcher,
+    private var writingJobDispatcher: CoroutineDispatcher = IODispatcher,
+    private var processingJobDispatcher: CoroutineDispatcher = Dispatchers.Default,
+    coroutineScope: CoroutineScope? = null,
+) : AbstractTransport() {
 
-    /**
-     * Configuration for [StdioServerTransport].
-     *
-     * @property source The input [Source] used to receive data.
-     * @property sink The output [Sink] used to send data.
-     * @property readBufferSize The buffer size for the read channel.
-     * @property readingJobDispatcher The [CoroutineDispatcher] used for reading jobs.
-     *      Defaults to [IODispatcher].
-     * @property writingJobDispatcher The [CoroutineDispatcher] used for writing jobs.
-     *      Defaults to [IODispatcher].
-     * @property processingJobDispatcher The [CoroutineDispatcher] used for processing jobs.
-     *      Defaults to [Dispatchers.Default].
-     * @property readChannelBufferSize The buffer size for the read channel.
-     * @property writeChannelBufferSize The buffer size for the write channel.
-     * @property coroutineScope The [CoroutineScope] used for managing coroutines.
-     */
-    @Suppress("LongParameterList")
-    public class Configuration internal constructor(
-        public var source: Source? = null,
-        public var sink: Sink? = null,
-        public var readBufferSize: Long = READ_BUFFER_SIZE,
-        public var readingJobDispatcher: CoroutineDispatcher = IODispatcher,
-        public var writingJobDispatcher: CoroutineDispatcher = IODispatcher,
-        public var processingJobDispatcher: CoroutineDispatcher = Dispatchers.Default,
-        public var readChannelBufferSize: Int = Channel.UNLIMITED,
-        public var writeChannelBufferSize: Int = Channel.UNLIMITED,
-        public var coroutineScope: CoroutineScope? = null,
-    )
-
-    private val source: Source
-    private val sink: Sink
-    private val processingJobDispatcher: CoroutineDispatcher
-    private val readingJobDispatcher: CoroutineDispatcher
-    private val writingJobDispatcher: CoroutineDispatcher
     private val scope: CoroutineScope
-    private val readBufferSize: Long
-    private val readChannel: Channel<ByteArray>
-    private val writeChannel: Channel<JSONRPCMessage>
+    private val sink: Sink
 
     init {
-        val config = Configuration().apply(block)
-        val input = requireNotNull(config.source) { "source is required" }
-        val output = requireNotNull(config.sink) { "sink is required" }
-        require(config.readBufferSize > 0) { "readBufferSize must be > 0" }
-
-        source = input
-        processingJobDispatcher = config.processingJobDispatcher
-        readingJobDispatcher = config.readingJobDispatcher
-        writingJobDispatcher = config.writingJobDispatcher
-        val parentJob = config.coroutineScope?.coroutineContext?.get(Job)
+        require(readBufferSize > 0) { "readBufferSize must be > 0" }
+        val parentJob = coroutineScope?.coroutineContext?.get(Job)
         scope = CoroutineScope(SupervisorJob(parentJob))
-        readBufferSize = config.readBufferSize
-        readChannel = Channel(config.readChannelBufferSize)
-        writeChannel = Channel(config.writeChannelBufferSize)
-        sink = output.buffered()
+        this.sink = sink.buffered()
     }
 
     /**
      * Creates a new instance of [StdioServerTransport]
      * with the given [inputStream] [Source] and [outputStream] [Sink].
      */
-    public constructor(inputStream: Source, outputStream: Sink) : this({
-        source = inputStream
-        sink = outputStream
-    })
+    public constructor(inputStream: Source, outputStream: Sink) : this(
+        source = inputStream,
+        sink = outputStream,
+    )
 
     private val logger = KotlinLogging.logger {}
     private val readBuffer = ReadBuffer()
     private val initialized: AtomicBoolean = AtomicBoolean(false)
-
-    @Volatile
-    private var readingJob: Job? = null
-
-    @Volatile
-    private var sendingJob: Job? = null
-
-    @Volatile
-    private var processingJob: Job? = null
 
     override suspend fun start() {
         if (!initialized.compareAndSet(expectedValue = false, newValue = true)) {
@@ -136,13 +100,13 @@ public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTra
         }
 
         // Launch a coroutine to read from stdin
-        readingJob = launchReadingJob()
+        launchReadingJob()
 
         // Launch a coroutine to process messages from readChannel
-        processingJob = launchProcessingJob()
+        launchProcessingJob()
 
         // Launch a coroutine to handle message sending
-        sendingJob = launchSendingJob()
+        launchSendingJob()
     }
 
     private fun launchReadingJob(): Job = scope.launch(readingJobDispatcher) {
@@ -186,7 +150,9 @@ public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTra
             _onError.invoke(e)
         }
     }.apply {
-        invokeOnCompletion { logJobCompletion("Processing", it) }
+        invokeOnCompletion { cause ->
+            logJobCompletion("Processing", cause)
+        }
     }
 
     private fun launchSendingJob(): Job = scope.launch(writingJobDispatcher) {
@@ -207,7 +173,7 @@ public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTra
         invokeOnCompletion { cause ->
             logJobCompletion("Message sending", cause)
             if (cause is CancellationException) {
-                readingJob?.cancel(cause)
+                readChannel.cancel(cause)
             }
         }
     }
@@ -255,22 +221,21 @@ public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTra
 
         withContext(NonCancellable) {
             writeChannel.close()
-            sendingJob?.cancelAndJoin()
 
             runCatching {
                 source.close()
             }.onFailure { logger.warn(it) { "Failed to close stdin" } }
 
-            readingJob?.cancel()
             readChannel.close()
-
-            processingJob?.cancelAndJoin()
 
             readBuffer.clear()
             runCatching {
                 sink.flush()
                 sink.close()
             }.onFailure { logger.warn(it) { "Failed to close stdout" } }
+
+            scope.cancel()
+            scope.coroutineContext[Job]?.join()
 
             invokeOnCloseCallback()
         }
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
@@ -8,7 +8,9 @@ import io.modelcontextprotocol.kotlin.sdk.shared.TransportSendOptions
 import io.modelcontextprotocol.kotlin.sdk.shared.serializeMessage
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCMessage
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
@@ -23,9 +25,9 @@ import kotlinx.io.Source
 import kotlinx.io.buffered
 import kotlinx.io.readByteArray
 import kotlinx.io.writeString
+import kotlin.concurrent.Volatile
 import kotlin.concurrent.atomics.AtomicBoolean
 import kotlin.concurrent.atomics.ExperimentalAtomicApi
-import kotlin.coroutines.CoroutineContext
 
 private const val READ_BUFFER_SIZE = 8192L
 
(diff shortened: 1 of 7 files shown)


kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
  *
- * @constructor Creates a new instance of [StdioServerTransport].
- * @param inputStream The input [Source] used to receive data.
- * @param outputStream The output [Sink] used to send data.
+ * Example:
+ * ```kotlin
+ * val transport = StdioServerTransport {
+ *   source = System.`in`.asInput(),
+ *   sink =  System.out.asSink(),
+ * }
```
devcrocod: It seems to me this is an incorrect example
as I remember, need to call buffered

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+ * ```
+ *
+ * @constructor Initializes the transport using the provided block for [Configuration].
+ * The configuration includes specifying the input and output streams, buffer sizes,
+ * and dispatchers for I/O and processing tasks and coroutine scope.
  */
 @OptIn(ExperimentalAtomicApi::class)
-public class StdioServerTransport(private val inputStream: Source, outputStream: Sink) : AbstractTransport() {
+public class StdioServerTransport(block: Configuration.() -> Unit) : AbstractTransport() {
```
devcrocod: Isn’t `Configuration` class used in streamable constructor?
Why is a builder used here then?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+     *      Defaults to [Dispatchers.Default].
+     * @property readChannelBufferSize The buffer size for the read channel.
+     * @property writeChannelBufferSize The buffer size for the write channel.
+     * @property coroutineScope The [CoroutineScope] used for managing coroutines.
+     */
+    @Suppress("LongParameterList")
+    public class Configuration internal constructor(
+        public var source: Source? = null,
+        public var sink: Sink? = null,
+        public var readBufferSize: Long = READ_BUFFER_SIZE,
```
devcrocod: This doesn’t need to be made public

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+        public var readingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var writingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var processingJobDispatcher: CoroutineDispatcher = Dispatchers.Default,
+        public var readChannelBufferSize: Int = Channel.UNLIMITED,
+        public var writeChannelBufferSize: Int = Channel.UNLIMITED,
```
devcrocod: This also shouldn’t be public

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+     * @property coroutineScope The [CoroutineScope] used for managing coroutines.
+     */
+    @Suppress("LongParameterList")
+    public class Configuration internal constructor(
+        public var source: Source? = null,
+        public var sink: Sink? = null,
+        public var readBufferSize: Long = READ_BUFFER_SIZE,
+        public var readingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var writingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var processingJobDispatcher: CoroutineDispatcher = Dispatchers.Default,
```
devcrocod: Why are there three dispatchers if one would be enough?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+        public var readChannelBufferSize: Int = Channel.UNLIMITED,
+        public var writeChannelBufferSize: Int = Channel.UNLIMITED,
+        public var coroutineScope: CoroutineScope? = null,
```
devcrocod: Why does coroutineScope need to be public?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
     private var sendingJob: Job? = null
-    private var processingJob: Job? = null
 
-    private val coroutineContext: CoroutineContext = IODispatcher + SupervisorJob()
-    private val scope = CoroutineScope(coroutineContext)
-    private val readChannel = Channel<ByteArray>(Channel.UNLIMITED)
-    private val writeChannel = Channel<JSONRPCMessage>(Channel.UNLIMITED)
-    private val outputSink = outputStream.buffered()
+    @Volatile
+    private var processingJob: Job? = null
```
devcrocod: Can we get rid of storing the Job fields and use a scope instead?
It seems to me that this would eliminate potential race conditions

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
```diff
@@ -69,6 +74,40 @@ class StdioServerTransportTest {
         printOutput = output.asSink().buffered()
     }
 
+    @OptIn(InternalAPI::class)
```
devcrocod: I think it’s better not to use InternalAPI from other libraries unless there’s a clear necessity

devcrocod (changes requested):
Please take a look at my comments

I’m not sure we need to extend the Configuration for stdio if it’s enough to just pass dispatcher for io

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: Creating buffered instance happens internally [here](https://github.com/modelcontextprotocol/kotlin-sdk/pull/584/changes#diff-bed24d2952344867fa13237822def589c348ac50426d31a736403c61d9e459a0R108)

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: What do you mean? StreamableHttpServerTransport uses different configuration class. This is `StdioServerTransport.Configuration`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: This is your opinion. There is a similar request for StreamableHttpTransport (#521). So, making buffer size customizable makes sense to me. Might be useful for performance tuning.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: One dispatcher is not enough. Here is why:
1. `processingJobDispatcher` for issue #573.
2. `readingJobDispatcher` and `writingJobDispatcher` should be configurable to run IO jobs on virtual threads. `Dispatchers.IO` uses platform threads by default.
3. I have considered having only one dispatcher for IO, but rejected this option for the sake of flexibility to implement the bulkhead pattern. This is crucial for high-load applications where read and write thread pools should be isolated.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: I disagree with this. Buffer sizes should be configurable for production systems. The default value of UNLIMITED is a clear performance issue (OOM risk). I am not changing it at this time to avoid unexpected changes for users but this is a defect for sure.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: #574

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: Let us address this issue within the scope of #518.

The `close()` method already contains a race condition which this PR does not exacerbate. The test is passing.

We should retain the current algorithm and pattern for the time being.

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransportTest.kt
kpavlov: I agree with you in general. This is for `clientDispatcher` and is not causing any issues here.

kpavlov (commented):
@devcrocod, thanks for your review! I’ve replied to your comments.
@e5l, I’d be grateful if you could also take a look.

kpavlov:
cc: @i1bro

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: I just noticed that commas are also used here

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: You are using two different API approaches to create transport

https://github.com/modelcontextprotocol/kotlin-sdk/blob/3e7ea39bf70556154699bdfe483ca049a16b99aa/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt#L75C1-L75C85

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: You are already making a change here, so instead of adding volatile, it could simply be removed jobs

devcrocod:
This PR introduces quite a few changes beyond addressing the specific issue
In my opinion, these changes are not ideal at the moment:
1. The resulting api is not very consistent and doesn’t look well designed
2. As noted in agents.md, the code should follow solid principles. In this case, it exposes too many internal implementation details
3. It makes it harder for us to change the internal implementation or change api

I think it would be better not to add many things to the public api without a request for them. We can always add this later without breaking compatibility, but removing or changing it afterwards would be much hard

kpavlov:
@devcrocod, where exactly this pr worsening existing design?
The PR is solving specific issues that were discovered by users. The rewriting of the transport is not the goal.
Could you please elaborate and provide constructive feedback?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+     * @property writingJobDispatcher The [CoroutineDispatcher] used for writing jobs.
+     *      Defaults to [IODispatcher].
+     * @property processingJobDispatcher The [CoroutineDispatcher] used for processing jobs.
+     *      Defaults to [Dispatchers.Default].
+     * @property readChannelBufferSize The buffer size for the read channel.
+     * @property writeChannelBufferSize The buffer size for the write channel.
+     * @property coroutineScope The [CoroutineScope] used for managing coroutines.
+     */
+    @Suppress("LongParameterList")
+    public class Configuration internal constructor(
```
nerzhulart: I'd not use such kind of builders. IMHO they look fancy but from use-site it's usually overengineering. I'd prefer put necessary parameters as parameters, like source and sink, as well as parent scope and io dispatcher (that's not presented in KMP). The default dispatcher can be overidden by passing a specific scope with the necessary dispatcher, but I'm sure the default one fits always. All the rest optional/default parmeters like buffer size I'd hide under StdioOptions data class with default instance if possible

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
nerzhulart: @kpavlov better do `val stdioDispatcher = ioDispatcher.limitedParallelism(2)` to have dedicated threads for your jobs. The don't consume the global IO dispatcher pool in this case. So you don't need to pass a dispatcher for each job (parameters should specify a kind of dispatcher for IO, but not every single dispatcher for every job)

nerzhulart: or you pass an IO dispatcher as a parameter and write a doc that at least 2 threads must be allocated in the dispatcher. So either a user passes a simple Dispatchers.IO, or also manually makes Dispatchers.IO.limitedParallelism(2)

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+        public var source: Source? = null,
+        public var sink: Sink? = null,
+        public var readBufferSize: Long = READ_BUFFER_SIZE,
+        public var readingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var writingJobDispatcher: CoroutineDispatcher = IODispatcher,
+        public var processingJobDispatcher: CoroutineDispatcher = Dispatchers.Default,
+        public var readChannelBufferSize: Int = Channel.UNLIMITED,
```
nerzhulart: when doing channel sizes configurable always specify where backpressure occurs (if occurs) and add tests for the limited cases

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
nerzhulart: I'd pass a scope as a constructor parameter

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -34,25 +36,99 @@ private const val READ_BUFFER_SIZE = 8192L
...
+    public constructor(inputStream: Source, outputStream: Sink) : this({
+        source = inputStream
+        sink = outputStream
+    })
 
     private val logger = KotlinLogging.logger {}
     private val readBuffer = ReadBuffer()
     private val initialized: AtomicBoolean = AtomicBoolean(false)
+
+    @Volatile
```
nerzhulart: I don't think you even need these fields for jobs. Just run an orchestration job in init and inside run all necessary jobs - read, write, processing together, then you may join them. In this case you get structured concurrency that joins in a single job. Less state (especially something like volatile) less problems

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -69,81 +145,71 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
...
+            throw e
+        } catch (e: Throwable) {
+            logger.error(e) { "Error writing to stdout" }
+            _onError.invoke(e)
         }
-        job.invokeOnCompletion { cause ->
+    }.apply {
+        invokeOnCompletion { cause ->
             logJobCompletion("Message sending", cause)
             if (cause is CancellationException) {
```
nerzhulart, line 175: I'd try to control cancellation via channels

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
```diff
@@ -192,7 +258,7 @@ public class StdioServerTransport(private val inputStream: Source, outputStream:
             sendingJob?.cancelAndJoin()
 
             runCatching {
-                inputStream.close()
+                source.close()
             }.onFailure { logger.warn(it) { "Failed to close stdin" } }
 
             readingJob?.cancel()
```
nerzhulart: when you hardly cancel the jobs they don't have a chance to terminate gracefully, I think better to close the channels and streams and handle semantic exceptions from them

nerzhulart (changes requested):

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: Accordingly, we need to settle on a single approach: either update/add a builder for streamable, or make the changes here for stdio

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: The approach you implemented does not follow from either the attached issue or the original one
Splitting into io and default, along with the ability to pass a value for io, fully addresses the request
agree with @nerzhulart

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: Of course, this is just my opinion, I'm not claiming it's truth
And I’m not saying that this is a bad change
At the same time, it locks api. As I mentioned in another comment, we can always add something later without trouble, but removing or changing it afterwards would be more problematic
If we want to pass such things through the main config, then that's ok

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: > The default value of UNLIMITED is a clear performance issue

That makes sense
At the same time, I'd still suggest keeping it private and changing this to `BUFFERED`. Leaving it as `UNLIMITED` seems unreasonable to me, sooner or later we'll likely have to change it anyway

Later on, we can add the ability to configure this parameter, which will be easier to do within an already agreed-upon API


I'd also suggest renaming it, since it's essentially a `capacity`

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: > The coroutine scope that's used in this transport is completely detached from the coroutine tree, which may cause leaks. Please add an optional CoroutineScope parameter

agree

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
devcrocod: I don't quite follow the reasoning here, if you're already doing a refactor, why not address this as well? Is there some constraint I’m missing?

devcrocod:
@kpavlov I replied to each comment again in case additional details were needed

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: This is public, because it's builder's field. It is copied to the private val field in constructor and the builder got distroyed

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StdioServerTransport.kt
kpavlov: If we want to configure backpressure, let's pass the entire Channel as a parameter.

kpavlov:
@e5l @devcrocod @nerzhulart I would appreciate it if you could share a link to the official Kotlin guide on writing idiomatic APIs. Until then, I am confused, as the answers appear to be based on personal preferences.

kpavlov:
Someone has shared this document with me: https://github.com/JetBrains/kotlin/blob/master/analysis/docs/contribution-guide/api-development.md. Also, I asked LLM to create [Kotlin api design guidelines](https://gist.github.com/kpavlov/5589747c37cac9833d1d06dbceb0206c)

devcrocod:
obsolete

Issue #545, opened 2026-02-19
split JsonRPCResponse


devcrocod:
[JSONRPCResponse](https://modelcontextprotocol.io/specification/2025-11-25/schema#jsonrpcresponse)

split on Result and Error

devcrocod:
check the serializer for an empty json rpc response: https://github.com/modelcontextprotocol/kotlin-sdk/pull/592#discussion_r2904361853

PR #592, opened 2026-03-07
fix: serialization empty response without id

### LINKED ISSUE #588: Client fails to deserialize JSON-RPC response without `id` field
- **Target:** client
- **Component:** sdk-client
- **Scenarios:** `initialize`

### Root cause

The conformance runner server sends a JSON-RPC response that contains a `result` field but no `id`
field. The `JSONRPCMessagePolymorphicSerializer` classifies it as a `JSONRPCResponse`, but `JSONRPCResponse.id` is
non-nullable and has no default, so kotlinx.serialization throws `MissingFieldException`. The
`StreamableHttpClientTransport` does not handle this case.

devcrocod:
fixes #588 

## How Has This Been Tested?
conformance test

## Breaking Changes
NaN

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/conformance-test/README.md
+++ b/conformance-test/README.md
@@ -110,7 +110,7 @@ Tests the conformance server against all server scenarios:
 
 ## Known SDK Limitations
 
-10 scenarios are expected to fail due to current SDK limitations (tracked in [
+9 scenarios are expected to fail due to current SDK limitations (tracked in [
 `conformance-baseline.yml`](conformance-baseline.yml).
 
 | Scenario                              | Suite  | Root Cause                                                                                                                                             |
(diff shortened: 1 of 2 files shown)


kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
```diff
@@ -371,21 +371,23 @@ internal object ServerResultPolymorphicSerializer :
...
+ * - "result" + "id" -> JSONRPCResponse
+ * - "result" -> JSONRPCEmptyMessage
  * - "method" + "id" -> JSONRPCRequest
  * - "method" -> JSONRPCNotification
  */
 internal object JSONRPCMessagePolymorphicSerializer :
     JsonContentPolymorphicSerializer<JSONRPCMessage>(JSONRPCMessage::class) {
     override fun selectDeserializer(element: JsonElement): DeserializationStrategy<JSONRPCMessage> {
-        val jsonObject = element.jsonObject
+        val jsonObj = element.jsonObject
```
e5l, line 382: We should check if we can fix it with schema changes. Could you log an issue to review for this case?

devcrocod: Yes, sure
we already have an issue for the json rpc response #545  
I added a comment about this case

PR #501, opened 2026-02-02
Detekt configuration

devcrocod:
- generate detekt config
- enable several rules
- disable ktlint backing property check

## Motivation and Context
Centralized control of detekt rules. Using detekt 2 configuration

## How Has This Been Tested?
locally

## Breaking Changes
-

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #596, opened 2026-03-10
feat: add SSE reconnection with retry support

### LINKED ISSUE #420: Implement SEP-1699: Support SSE Polling via Server-Side Disconnect
This is a tracking issue for implementation of [SEP-1699](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699).

## Summary

This SEP proposes changes to the Streamable HTTP transport to mitigate issues with long-running connections and improve resumability. Servers must immediately send an SSE event with an id and empty data string to prime clients for reconnection, and may disconnect at will after sending an event ID. Servers should send a retry field to prevent excessive reconnection attempts, and clients must respect this retry field. This allows servers to avoid maintaining potentially long-running connections while computing results.

The Kotlin SDK's current SSE implementation in `SSEServerTransport` and `StreamableHttpClientTransport` does not fully support this pattern. The server transport does not send initial priming events with IDs, does not include retry fields in SSE events, and maintains connections until responses are sent. The client transport already handles Last-Event-ID for resumption and supports reconnection via Ktor's SSE client, but does not explicitly handle or respect retry fields from the server. This implementation requires: server-side changes to send initial SSE events with IDs and empty data upon connection, support for including retry fields in SSE events, ability to disconnect before sending JSON-RPC responses (after sending event ID), and client-side changes to parse and respect retry fields when reconnecting. The changes should maintain backward compatibility with clients/servers that don't support this pattern.

### LINKED ISSUE #590: Client does not respect SSE `retry` field or send `Last-Event-ID` on reconnection
- **Target:** client
- **Component:** sdk-client
- **Scenarios:** `sse-retry`

### Root cause

When the server closes an SSE stream to trigger reconnection, `StreamableHttpClientTransport` reconnects immediately instead of waiting the duration specified by the SSE `retry` field (e.g. 500 ms). It also does not send the `Last-Event-ID` header on the reconnection GET request. The Ktor SSE client manages reconnection internally and the transport does not propagate the `retry` timing or last event ID into its own reconnection logic

Client-side subset of #420

devcrocod:
Add configurable SSE reconnection with exponential backoff and server-driven retry delays to `StreamableHttpClientTransport`

closes #590
closes #420 

## How Has This Been Tested?
New unit tests and pass conformance test 

## Breaking Changes
old constructors are Deprecated
`close` no longer calls `terminateSession`

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
@@ -41,8 +41,6 @@ import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
-import kotlin.concurrent.Volatile
-import kotlin.concurrent.atomics.ExperimentalAtomicApi
 import kotlin.math.pow
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
@@ -58,12 +56,17 @@ private const val MCP_RESUMPTION_TOKEN_HEADER = "Last-Event-ID"
 public class StreamableHttpError(public val code: Int? = null, message: String? = null) :
     Exception("Streamable HTTP error: $message")
 
+private sealed interface ConnectResult {
+    data class Success(val session: ClientSSESession) : ConnectResult
+    data object NonRetryable : ConnectResult
+    data object Failed : ConnectResult
+}
+
 /**
  * Client transport for Streamable HTTP: this implements the MCP Streamable HTTP transport specification.
  * It will connect to a server using HTTP POST for sending messages and HTTP GET with Server-Sent Events
  * for receiving messages.
  */
-@OptIn(ExperimentalAtomicApi::class)
 @Suppress("TooManyFunctions")
 public class StreamableHttpClientTransport(
     private val client: HttpClient,
@@ -94,20 +97,17 @@ public class StreamableHttpClientTransport(
         private set
     public var protocolVersion: String? = null
 
-    private var sseSession: ClientSSESession? = null
     private var sseJob: Job? = null
-    private var reconnectJob: Job? = null
 
     private val scope by lazy { CoroutineScope(SupervisorJob() + Dispatchers.Default) }
 
-    @Volatile
-    private var lastEventId: String? = null
-
-    @Volatile
-    private var serverRetryDelay: Duration? = null
-
-    /** Result of SSE stream collection. Reconnect when [hasPrimingEvent] is true and [receivedResponse] is false. */
-    private data class SseStreamResult(val hasPrimingEvent: Boolean, val receivedResponse: Boolean)
+    /** Result of an SSE stream collection. Reconnect when [hasPrimingEvent] is true and [receivedResponse] is false. */
+    private data class SseStreamResult(
+        val hasPrimingEvent: Boolean,
+        val receivedResponse: Boolean,
+        val lastEventId: String? = null,
+        val serverRetryDelay: Duration? = null,
+    )
 
     override suspend fun initialize() {
         logger.debug { "Client transport is starting..." }
@@ -165,26 +165,15 @@ public class StreamableHttpClientTransport(
             }
 
             ContentType.Text.EventStream -> {
-                val result = handleInlineSse(
-                    response,
-                    onResumptionToken = options?.onResumptionToken,
-                    replayMessageId = if (message is JSONRPCRequest) message.id else null,
-                )
+                val replayMessageId = if (message is JSONRPCRequest) message.id else null
+                val result = handleInlineSse(response, replayMessageId, options?.onResumptionToken)
                 if (result.hasPrimingEvent && !result.receivedResponse) {
-                    reconnectJob = scope.launch {
-                        try {
-                            serverRetryDelay?.let { delay(it) }
-                            startSseSession(
-                                resumptionToken = lastEventId,
-                                replayMessageId = if (message is JSONRPCRequest) message.id else null,
-                                onResumptionToken = options?.onResumptionToken,
-                            )
-                        } catch (e: CancellationException) {
-                            throw e
-                        } catch (e: Exception) {
-                            logger.debug { "POST-to-GET SSE reconnection failed: ${e.message}" }
-                        }
-                    }
+                    startSseSession(
+                        resumptionToken = result.lastEventId,
+                        replayMessageId = replayMessageId,
+                        onResumptionToken = options?.onResumptionToken,
+                        initialServerRetryDelay = result.serverRetryDelay,
+                    )
                 }
             }
 
@@ -218,13 +207,7 @@ public class StreamableHttpClientTransport(
 
     override suspend fun closeResources() {
         logger.debug { "Client transport closing." }
-
-        // Try to terminate session if we have one
-        terminateSession()
-
-        sseSession?.cancel()
         sseJob?.cancelAndJoin()
-        reconnectJob?.cancelAndJoin()
         scope.cancel()
     }
 
@@ -251,78 +234,94 @@ public class StreamableHttpClientTransport(
         }
 
         sessionId = null
-        lastEventId = null
         logger.debug { "Session terminated successfully" }
     }
 
-    private suspend fun startSseSession(
+    private fun startSseSession(
         resumptionToken: String? = null,
         replayMessageId: RequestId? = null,
         onResumptionToken: ((String) -> Unit)? = null,
+        initialServerRetryDelay: Duration? = null,
     ) {
-        sseSession?.cancel()
-        sseJob?.cancelAndJoin()
+        // Cancel-and-replace: cancel() signals the previous job, join() inside
+        // the new coroutine ensures it completes before we start collecting.
+        // This is intentionally non-suspend to avoid blocking performSend.
+        val previousJob = sseJob
+        previousJob?.cancel()
+        sseJob = scope.launch(CoroutineName("StreamableHttpTransport.collect#${hashCode()}")) {
+            previousJob?.join()
+            var lastEventId = resumptionToken
+            var serverRetryDelay = initialServerRetryDelay
+            var attempt = 0
+            var needsDelay = initialServerRetryDelay != null
 
+            @Suppress("LoopWithTooManyJumpStatements")
+            while (isActive) {
+                // Delay before (re)connection: skip only for first fresh SSE connection
+                if (needsDelay) {
+                    delay(getNextReconnectionDelay(attempt, serverRetryDelay))
+                }
+                needsDelay = true
+
+                // Connect
+                val session = when (val cr = connectSse(lastEventId)) {
+                    is ConnectResult.Success -> {
+                        attempt = 0
+                        cr.session
+                    }
+
+                    ConnectResult.NonRetryable -> return@launch
+
+                    ConnectResult.Failed -> {
+                        // Give up after maxRetries consecutive failed connection attempts
+                        if (++attempt >= reconnectionOptions.maxRetries) {
+                            _onError(StreamableHttpError(null, "Maximum reconnection attempts exceeded"))
+                            return@launch
+                        }
+                        continue
+                    }
+                }
+
+                // Collect
+                val result = collectSse(session, replayMessageId, onResumptionToken)
+                lastEventId = result.lastEventId ?: lastEventId
+                serverRetryDelay = result.serverRetryDelay ?: serverRetryDelay
+                if (result.receivedResponse) break
+            }
+        }
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private suspend fun connectSse(lastEventId: String?): ConnectResult {
         logger.debug { "Client attempting to start SSE session at url: $url" }
-        try {
-            sseSession = client.sseSession(
-                urlString = url,
-                showRetryEvents = true,
-            ) {
+        return try {
+            val session = client.sseSession(urlString = url, showRetryEvents = true) {
                 method = HttpMethod.Get
                 applyCommonHeaders(this)
-                // sseSession will add ContentType.Text.EventStream automatically
                 accept(ContentType.Application.Json)
-                (resumptionToken ?: lastEventId)?.let { headers.append(MCP_RESUMPTION_TOKEN_HEADER, it) }
+                lastEventId?.let { headers.append(MCP_RESUMPTION_TOKEN_HEADER, it) }
                 requestBuilder()
             }
             logger.debug { "Client SSE session started successfully." }
-        } catch (e: SSEClientException) {
-            if (isNonRetryableSseError(e)) return
+            ConnectResult.Success(session)
+        } catch (e: CancellationException) {
             throw e
-        }
-
-        sseJob = scope.launch(CoroutineName("StreamableHttpTransport.collect#${hashCode()}")) {
-            @Suppress("LoopWithTooManyJumpStatements")
-            while (isActive) {
-                val result = sseSession?.let { collectSse(it, replayMessageId, onResumptionToken) } ?: break
-                if (result.receivedResponse) break
-                if (!reconnectSseSession()) {
-                    _onError(StreamableHttpError(null, "Maximum reconnection attempts exceeded"))
-                    break
-                }
+        } catch (e: SSEClientException) {
+            if (isNonRetryableSseError(e)) {
+                ConnectResult.NonRetryable
+            } else {
+                logger.debug { "SSE connection failed: ${e.message}" }
+                ConnectResult.Failed
             }
+        } catch (e: Exception) {
+            logger.debug { "SSE connection failed: ${e.message}" }
+            ConnectResult.Failed
         }
     }
 
-    @Suppress("ReturnCount", "TooGenericExceptionCaught")
-    private suspend fun reconnectSseSession(): Boolean {
-        for (attempt in 0 until reconnectionOptions.maxRetries) {
-            delay(getNextReconnectionDelay(attempt))
-            try {
-                sseSession = client.sseSession(
-                    urlString = url,
-                    showRetryEvents = true,
-                ) {
-                    method = HttpMethod.Get
-                    applyCommonHeaders(this)
-                    accept(ContentType.Application.Json)
-                    lastEventId?.let { headers.append(MCP_RESUMPTION_TOKEN_HEADER, it) }
-                    requestBuilder()
-                }
-                return true
-            } catch (e: CancellationException) {
-                throw e
-            } catch (e: SSEClientException) {
-                if (isNonRetryableSseError(e)) return false
-            } catch (e: Exception) {
-                logger.debug { "SSE reconnection attempt $attempt failed: ${e.message}" }
-            }
-        }
-        return false
-    }
-
-    private fun getNextReconnectionDelay(attempt: Int): Duration {
+    private fun getNextReconnectionDelay(attempt: Int, serverRetryDelay: Duration?): Duration {
+        // Per SSE specification, the server-sent `retry` field sets the reconnection time
+        // for all subsequent attempts, taking priority over exponential backoff.
         serverRetryDelay?.let { return it }
         val delay = reconnectionOptions.initialReconnectionDelay *
             reconnectionOptions.reconnectionDelayMultiplier.pow(attempt)
@@ -367,11 +366,13 @@ public class StreamableHttpClientTransport(
     ): SseStreamResult {
         var hasPrimingEvent = false
         var receivedResponse = false
+        var localLastEventId: String? = null
+        var localServerRetryDelay: Duration? = null
         try {
             session.incoming.collect { event ->
-                event.retry?.let { serverRetryDelay = it.milliseconds }
+                event.retry?.let { localServerRetryDelay = it.milliseconds }
                 event.id?.let {
-                    lastEventId = it
+                    localLastEventId = it
                     hasPrimingEvent = true
                     onResumptionToken?.invoke(it)
                 }
@@ -399,7 +400,7 @@ public class StreamableHttpClientTransport(
         } catch (t: Throwable) {
             _onError(t)
         }
-        return SseStreamResult(hasPrimingEvent, receivedResponse)
+        return SseStreamResult(hasPrimingEvent, receivedResponse, localLastEventId, localServerRetryDelay)
     }
 
     @Suppress("CyclomaticComplexMethod")
@@ -413,13 +414,15 @@ public class StreamableHttpClientTransport(
 
         var hasPrimingEvent = false
         var receivedResponse = false
+        var localLastEventId: String? = null
+        var localServerRetryDelay: Duration? = null
         val sb = StringBuilder()
         var id: String? = null
         var eventName: String? = null
 
         suspend fun dispatch(id: String?, eventName: String?, data: String) {
             id?.let {
-                lastEventId = it
+                localLastEventId = it
                 hasPrimingEvent = true
                 onResumptionToken?.invoke(it)
             }
@@ -466,10 +469,10 @@ public class StreamableHttpClientTransport(
                 line.startsWith("data:") -> sb.append(line.substringAfter("data:").trim())
 
                 line.startsWith("retry:") -> line.substringAfter("retry:").trim().toLongOrNull()?.let {
-                    serverRetryDelay = it.milliseconds
+                    localServerRetryDelay = it.milliseconds
                 }
             }
         }
-        return SseStreamResult(hasPrimingEvent, receivedResponse)
+        return SseStreamResult(hasPrimingEvent, receivedResponse, localLastEventId, localServerRetryDelay)
     }
 }
--- a/kotlin-sdk-client/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTest.kt
+++ b/kotlin-sdk-client/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTest.kt
@@ -1,6 +1,7 @@
 package io.modelcontextprotocol.kotlin.sdk.client
 
 import io.kotest.matchers.collections.shouldContain
+import io.kotest.matchers.shouldBe
 import io.ktor.http.ContentType
 import io.ktor.http.HttpMethod
 import io.ktor.http.HttpStatusCode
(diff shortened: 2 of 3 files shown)


kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -58,13 +64,28 @@ public class StreamableHttpError(public val code: Int? = null, message: String?
...
+        "Use constructor with ReconnectionOptions",
+        replaceWith = ReplaceWith(
+            "StreamableHttpClientTransport(client, url, " +
+                "ReconnectionOptions(initialReconnectionDelay = reconnectionTime ?: 1.seconds), requestBuilder)",
+        ),
+    )
+    public constructor(
+        client: HttpClient,
+        url: String,
+        reconnectionTime: Duration?,
```
kpavlov, line 90: Previously, it had a `null` as a default value. This is incompatible change.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -133,18 +161,34 @@ public class StreamableHttpClientTransport(
...
+                            startSseSession(
+                                resumptionToken = lastEventId,
+                                replayMessageId = if (message is JSONRPCRequest) message.id else null,
+                                onResumptionToken = options?.onResumptionToken,
+                            )
+                        } catch (e: Exception) {
+                            logger.debug { "POST-to-GET SSE reconnection failed: ${e.message}" }
+                        }
+                    }
+                }
```
kpavlov: consider extracting method for clarity

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @property initialReconnectionDelay The initial delay before the first reconnection attempt.
+ * @property maxReconnectionDelay The maximum delay between reconnection attempts.
+ * @property reconnectionDelayGrowFactor The factor by which the delay grows on each attempt.
+ * @property maxRetries The maximum number of reconnection attempts per disconnect.
+ */
+public class ReconnectionOptions(
+    public val initialReconnectionDelay: Duration = 1.seconds,
+    public val maxReconnectionDelay: Duration = 30.seconds,
+    public val reconnectionDelayGrowFactor: Double = 1.5,
+    public val maxRetries: Int = 2,
```
kpavlov, line 18: It is also advantageous to have a jitter.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -133,18 +161,34 @@ public class StreamableHttpClientTransport(
...
+                if (result.hasPrimingEvent && !result.receivedResponse) {
+                    scope.launch {
+                        try {
+                            serverRetryDelay?.let { delay(it) }
+                            startSseSession(
+                                resumptionToken = lastEventId,
+                                replayMessageId = if (message is JSONRPCRequest) message.id else null,
+                                onResumptionToken = options?.onResumptionToken,
+                            )
+                        } catch (e: Exception) {
```
kpavlov: catch (e: Exception) silently swallows CancellationException

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -228,28 +272,73 @@ public class StreamableHttpClientTransport(
...
+                ) {
+                    method = HttpMethod.Get
+                    applyCommonHeaders(this)
+                    accept(ContentType.Application.Json)
+                    lastEventId?.let { headers.append(MCP_RESUMPTION_TOKEN_HEADER, it) }
+                    requestBuilder()
+                }
+                return true
+            } catch (e: SSEClientException) {
+                if (isNonRetryableSseError(e)) return false
```
kpavlov: Non-SSEClientException errors in reconnectSseSession are silently lost.

Ktor's sseSession() throws SSEClientException for HTTP-level errors, but ordinary network failures — IOException, UnresolvedAddressException, connection timeouts — are thrown as-is, uncaught here. They propagate out of the sseJob coroutine body, which has no try/catch.

Under a SupervisorJob, the escaping exception does not cancel the scope — it is sent to the uncaught exception handler (stderr on JVM). _onError is never called. Reconnect failures from network errors are silently lost from the caller's perspective.
  

  Fix: Wrap the sseJob body:                                                                                                                                 
   
  ```kotlin
sseJob = scope.launch(...) {                                                                                                                               
      try {                                                                                                                                                  
          while (isActive) { ... }                                                                                                                           
      } catch (_: CancellationException) {                                                                                                                   
          // normal cancellation                                                                                                                             
      } catch (t: Throwable) {                                                                                                                             
          _onError(t)                                                                                                                                        
      }
  }  
```                                                                                                                                                        
                                                                                                                                                           
  Alternatively, widen the catch in reconnectSseSession to cover all retryable Exception types (excluding CancellationException).

File: kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt [hunk 6]
```diff
@@ -133,18 +161,34 @@ public class StreamableHttpClientTransport(
...
-                replayMessageId = if (message is JSONRPCRequest) message.id else null,
-            )
+            ContentType.Text.EventStream -> {
+                val result = handleInlineSse(
+                    response,
+                    onResumptionToken = options?.onResumptionToken,
+                    replayMessageId = if (message is JSONRPCRequest) message.id else null,
+                )
+                if (result.hasPrimingEvent && !result.receivedResponse) {
+                    scope.launch {
```
kpavlov: The Job returned by scope.launch { startSseSession(...) } is discarded. closeResources() does:                                                             
                                                                                                                                                             
```kotlin
sseSession?.cancel()                                                                                                                                       
sseJob?.cancelAndJoin()   // only tracks the job from startSseSession's internals                                                                        
scope.cancel()            // cancels children but does not join them   
```                                                                                    
   
There is no join on the untracked coroutine. If close() is called while this coroutine is mid-flight inside startSseSession() — which internally does sseJob?.cancelAndJoin() and sets sseJob to a new value — the new sseJob reference may race with closeResources()'s own sseJob?.cancelAndJoin(). The transport can appear closed while a reconnect coroutine is still active.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -58,13 +64,28 @@ public class StreamableHttpError(public val code: Int? = null, message: String?
...
-    private val reconnectionTime: Duration? = null,
+    private val reconnectionOptions: ReconnectionOptions = ReconnectionOptions(),
     private val requestBuilder: HttpRequestBuilder.() -> Unit = {},
 ) : AbstractClientTransport() {
 
+    @Deprecated(
+        "Use constructor with ReconnectionOptions",
+        replaceWith = ReplaceWith(
+            "StreamableHttpClientTransport(client, url, " +
+                "ReconnectionOptions(initialReconnectionDelay = reconnectionTime ?: 1.seconds), requestBuilder)",
```
kpavlov, line 82: ReplaceWith has an imports vararg. Without it, the IDE's quick-fix inserts the replacement text but does not add the required imports. The result is unresolvable ReconnectionOptions and 1.seconds in the caller's file.

Use GQN kotlin.time.Duration.Companion.seconds and io.modelcontextprotocol.kotlin.sdk.client.ReconnectionOptions

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @property initialReconnectionDelay The initial delay before the first reconnection attempt.
+ * @property maxReconnectionDelay The maximum delay between reconnection attempts.
+ * @property reconnectionDelayGrowFactor The factor by which the delay grows on each attempt.
+ * @property maxRetries The maximum number of reconnection attempts per disconnect.
+ */
+public class ReconnectionOptions(
+    public val initialReconnectionDelay: Duration = 1.seconds,
+    public val maxReconnectionDelay: Duration = 30.seconds,
+    public val reconnectionDelayGrowFactor: Double = 1.5,
```
kpavlov:
```suggestion
    public val delayMultiplier: Double = 1.5,
```

"Grow" is not standard terminology; the conventional term is "growth factor" or "multiplier". All comparable Kotlin retry libraries (ktor-client, resilience4j-kotlin) use "multiplier" or "factor" — never "growFactor".

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -76,16 +97,23 @@ public class StreamableHttpClientTransport(
 
     private val scope by lazy { CoroutineScope(SupervisorJob() + Dispatchers.Default) }
 
+    @Volatile
```
kpavlov: @devcrocod, why are you adding `@Volatile` here? This [not recommended here](https://github.com/modelcontextprotocol/kotlin-sdk/pull/584/changes/45e780b8d7c4561c3f6a1201ef74a5286127050c#r2885804215).

And still, there is a correctness problem: sseSession written from background coroutine without `@Volatile`

serverRetryDelay and lastEventId both received `@Volatile` in this PR. sseSession did not, even though it is now also written from a background coroutine (the sseJob loop via reconnectSseSession). closeResources() reads sseSession before joining sseJob, so there is no happens-before:                         
                                                                                                                                                             
```kotlin
sseSession?.cancel()       // may read stale reference                                                                                                   
sseJob?.cancelAndJoin()    // happens-before only established after this
```
                                                                                                                                                             
Consider adding `@Volatile` to sseSession for consistency and correctness.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: The new constructor has a default parameter
if we add it to the old one, it will cause an overload resolution ambiguity

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: scope.launch is used here. Creating a new method would require passing a lot of things around. It would only increase the complexity just to extract a dozen lines

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
devcrocod: good idea, will consider in follow-up

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
devcrocod: I didn't find where ktor-client uses "multiplier" or "factor"
Nevertheless, I’m thinking of changing it to `reconnectionDelayMultiplier`

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: idea resolves the imports, I checked
I can add it, that's not a problem

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: Agreed, I missed @Volatile on sseSession. I didn't originally plan to refactor this transport in this PR, but I went ahead and refactored it

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -52,31 +56,58 @@ private const val MCP_RESUMPTION_TOKEN_HEADER = "Last-Event-ID"
...
+    ) : this(client, url, ReconnectionOptions(initialReconnectionDelay = reconnectionTime ?: 1.seconds), requestBuilder)
+
     override val logger: KLogger = KotlinLogging.logger {}
 
     public var sessionId: String? = null
         private set
     public var protocolVersion: String? = null
 
-    private var sseSession: ClientSSESession? = null
     private var sseJob: Job? = null
```
kpavlov, line 100:
```suggestion
     private val sseJobRef = AtomicReference<Job?>(null)
```

sseJob is written in startSseSession() — called from the caller's coroutine via performSend() — and read in closeResources() — called from whatever coroutine invokes close(). These are independent coroutines, potentially running on different threads. Without `@Volatile` or an AtomicReference, the JVM memory model does not guarantee the write in startSseSession is visible to the thread executing closeResources.                                                            
                            
The consequence: sseJob?.cancelAndJoin() in closeResources may operate on a stale reference. scope.cancel() serves as a safety net (all children are cancelled regardless), but cancelAndJoin() — which is meant to wait for the SSE coroutine to drain — may be skipped on the wrong reference. close() can return while the SSE coroutine is still running.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        if (reconnectionDelayMultiplier != other.reconnectionDelayMultiplier) return false
+        if (maxRetries != other.maxRetries) return false
+        if (initialReconnectionDelay != other.initialReconnectionDelay) return false
+        if (maxReconnectionDelay != other.maxReconnectionDelay) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
```
kpavlov, line 34: the hashCode() ordering doesn't match the equals() ordering (reconnectionDelayMultiplier first vs. last), which is a style inconsistency even if not a correctness bug.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
kpavlov, line 1: 1. CancellationException swallowed in `collectSse` — breaks cancellation contract

2. Unowned CoroutineScope — no structured lifecycle guarantee

```kotlin
  private val scope by lazy { CoroutineScope(SupervisorJob() + Dispatchers.Default) }
```

This scope has no parent Job and is not linked to any caller's scope. It is manually cancelled in closeResources(). The lazy initialization means the scope isn't created until startSseSession() is first invoked, which avoids a leak if the transport is never actually used. But if it is initialized and close() is never called (e.g., caller drops the reference), the scope and all its children leak silently.

The fundamental issue: callers can't inject a lifecycle via a parent scope. The transport creates its own root, which is the root cause of needing explicit scope.cancel() in teardown.

This is a structural constraint that's hard to fix without changing the public API, but it should be documented explicitly so future contributors understand why closeResources() is not optional.

kotlin-sdk-client/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTest.kt
kpavlov, line 1: ## testTerminateSession is a false positive

  Location: StreamableHttpClientTransportTest.kt:185–202

```kotlin
  val transport = createTransport { request ->
      assertEquals(HttpMethod.Delete, request.method)
      assertEquals("test-session-id", request.headers["mcp-session-id"])
      respond(content = "", status = HttpStatusCode.OK)
  }

  transport.start()
  transport.terminateSession()   // sessionId == null → returns immediately

  assertNull(transport.sessionId)  // trivially true: was null before the call
```

  terminateSession() returns early on sessionId == null. The MockRequestHandler is never invoked. The assertions inside the handler never execute.
  assertNull(transport.sessionId) passes trivially. The test provides zero coverage of the DELETE-request path and gives false confidence.

  The same applies to testTerminateSessionHandle405.

  The root cause is that sessionId now has private set, so tests can't seed it directly. The fix is to drive a full handshake first (POST initialize → receive
  mcp-session-id header → then call terminateSession()), or to add a test-only constructor parameter. As written, these tests should be marked @Ignore with a
  note explaining what they actually cover.

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: scope.cancel() in closeResources will cancel all child coroutines

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
devcrocod: ?

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
devcrocod: > CancellationException swallowed in collectSse — breaks cancellation contract

nope
CancellationException is handled deliberately rather than accidentally. collectSse is a terminal function that returns a result. After catch, the coroutine checks isActive and exits properly, so this does not violate the contract

> Unowned CoroutineScope — no structured lifecycle guarantee

How is this comment related to this PR?

devcrocod:
@kpavlov please, if you use ai review, check the comments before posting them

 I don't see a problem with merging this into conformance-tests

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
kpavlov: Not related, just observation. But your explanation (not sure if it's correct approach) reveals non-trivial logic that needs to be simplified for better maintainability.

kpavlov:
@devcrocod, GH checks did not run for this PR. This is good enough reason for me.

kpavlov:
Good point about checking AI comments. I do my best to review the comments before sharing them.

In several recent PRs, AI quickly identified gaps and some non-trivial logical issues that could have been addressed earlier. @devrocod, could you please run an AI self-review before requesting human review in the future? This would likely make the process smoother for everyone.

devcrocod:
@kpavlov 
> GH checks did not run for this PR. This is good enough reason for me

https://github.com/modelcontextprotocol/kotlin-sdk/actions/runs/22949664879

> @devrocod, could you please run an AI self-review before requesting human review in the future? 

I didn’t request a review from you on this PR. In fact, I hadn’t requested one from anyone initially because I thought I would refine it further. And yes, I use different tools

e5l (approved):

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
```diff
@@ -0,0 +1,44 @@
...
+ * @property reconnectionDelayMultiplier The factor by which the delay grows on each attempt.
+ * @property maxRetries The maximum number of reconnection attempts per disconnect.
+ */
+public class ReconnectionOptions(
+    public val initialReconnectionDelay: Duration = 1.seconds,
+    public val maxReconnectionDelay: Duration = 30.seconds,
+    public val reconnectionDelayMultiplier: Double = 1.5,
+    public val maxRetries: Int = 2,
+) {
+    override fun equals(other: Any?): Boolean {
```
nerzhulart, line 20: why don't use data class for this? to avoid manual hashCode/equals, and also it gives you the pattern options.copy([only specific fields])

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -85,7 +116,7 @@ public class StreamableHttpClientTransport(
     /**
      * Sends a single message with optional resumption support
      */
-    @Suppress("ReturnCount", "CyclomaticComplexMethod")
+    @Suppress("ReturnCount", "CyclomaticComplexMethod", "LongMethod", "TooGenericExceptionCaught", "ThrowsCount")
```
nerzhulart, line 119: пацаны, вот серьезно, зачем вы нацепили этих говнокодерских линтеров?) как будто на джаваскрипте пишем)

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ReconnectionOptions.kt
devcrocod: Data classes are indeed more convenient

the problem is that any changes to them can introduce binary incompatibility in the future
https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
```diff
@@ -201,55 +234,120 @@ public class StreamableHttpClientTransport(
...
+        initialServerRetryDelay: Duration? = null,
     ) {
-        sseSession?.cancel()
-        sseJob?.cancelAndJoin()
+        // Cancel-and-replace: cancel() signals the previous job, join() inside
+        // the new coroutine ensures it completes before we start collecting.
+        // This is intentionally non-suspend to avoid blocking performSend.
+        val previousJob = sseJob
+        previousJob?.cancel()
+        sseJob = scope.launch(CoroutineName("StreamableHttpTransport.collect#${hashCode()}")) {
```
nerzhulart, line 251: Is it possible that startSseSession can be invoked in parallel? If so the code can race: you may get a abandoned job that works in parallel to the main saved to sseJob

devcrocod: Yes, I considered this scenario
Even with parallel calls, it should work correctly. The new job will wait for the previous one to finish

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StreamableHttpClientTransport.kt
kpavlov: This serves as a reminder of technical debt to address. "CyclomaticComplexMethod" and "LongMethod" are particularly valuable.

devcrocod: > This serves as a reminder of technical debt to address. 

Baseline is technical debt

In my opinion, the source code should not contain Suppress annotations for third-party tools. This tends to clutter the codebase and can make long-term maintenance harder, especially if we later migrate to different tooling

Detekt is a useful plugin, but I currently have a couple of concerns about using it in this project

First, every project or product has its own lifecycle. At the current stage of the kotlin-sdk development, this plugin feels somewhat excessive. It slows down development without providing much value at the moment, and the codebase becomes filled with Suppress annotations. It may also discourage external contributors from participating

Second, I'm not convinced that relying purely on the default rules is the right approach. It’s challenging to create large, universal rule sets that work well for every type of development: frameworks, libraries, enterprise applications, compose apps, etc. With the default setup, it’s also not very clear which rules are enabled or disabled and what their configured values are

For example, it was mentioned that LongMethod is particularly valuable. However, do we know what exactly counts as "long"? How many lines? In a previous version it was 80 lines, and now it’s 60. Changes like this raise questions

The Detekt maintainers understand that rules often need to be adapted for a specific project and team. That’s why they allow generating the default configuration and customizing it. Additionally, some rules that could be useful for an SDK are disabled by default

I tried to address this in PR #501 by introducing an explicit configuration. However, for reasons I still don't fully understand, this change was later removed in #512. My intention was to keep a clear configuration file that explicitly shows which rules are enabled, what values they use, and allows us to track changes over time, whether they come from our adjustments or from detekt updates. Explicit configuration also helps avoid issues where the gradle plugin and the idea plugin behave differently. I also enabled some rules and relaxed others to better fit our sdk use case

I'm honestly a bit disappointed that these concerns seem to be ignored or dismissed. When it comes to team tooling decisions, I would really appreciate my perspective being considered as part of the team discussion. For example, detekt itself was initially introduced bypassing the maintainers and merged without their approval in #493

PR #585, opened 2026-03-07
feat!: add mcp conformance test infra (#585); add SSE reconnection with retry support (#596)

devcrocod:
Adds a comprehensive conformance test suite for the Kotlin MCP SDK, covering core protocol operations, tool calls, elicitation, resources, prompts, and 20 OAuth/auth scenarios

- Conformance server and client implementations
- OAuth/auth test scenarios: JWT, authorization code flow, client credentials, PKCE, scope handling, cross-app access, client registration
- CI workflow
- Baseline file for tracking expected failures
- Shell script

fixes:
- #592
- #593
- #596

## Remaining known failures (tracked issues, will be fixed directly in `main`)

- [x] `tools-call-with-logging`, `tools-call-with-progress`, `tools-call-sampling`, `tools-call-elicitation`,
`elicitation-sep1034-defaults`-  see #599,
- [x] `elicitation-sep1330-enums` - #587 #600
- [x] `initialize` - #588 
- [x] `tools_call`, `auth/scope-step-up`, `auth/scope-retry-limit` - #589 
- [ ] `elicitation-sep1034-client-defaults` - #414 
- [x] `sse-retry` - #590 
- [ ] `resources-templates-read` - #591 

## Breaking Changes
from #596 
- `StreamableHttpClientTransport` and `mcpStreamableHttp`/`mcpStreamableHttpTransport`: old constructors accepting `Duration` timeout are now `@Deprecated` — use the new overloads with `ReconnectionOptions` instead
- `StreamableHttpClientTransport.close()` no longer calls `terminateSession()` automatically

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #599, opened 2026-03-12
fix(server): notifications not delivered over Streamable HTTP (#587)

### LINKED ISSUE #587: Server-initiated notifications from tool handlers are not delivered to the client
- **Target:** server
- **Component:** sdk-server
- **Scenarios:** `tools-call-with-logging`, `tools-call-with-progress`, `tools-call-sampling`, `tools-call-elicitation`,
  `elicitation-sep1034-defaults`, `elicitation-sep1330-enums`

### Root cause
Notifications and requests sent from a tool handler (`notification()`, `sendLoggingMessage()`, `createMessage()`) have no `relatedRequestId`. The Streamable HTTP transport therefore routes them to the standalone SSE stream instead of the SSE stream associated with the originating `tools/call` POST request. The conformance runner only reads from the request-specific stream, so it never receives these messages (notifications are lost, requests time out)

kpavlov:
## Changes

- fix(server): keep SSE connection open until explicitly cancelled
- conformance: add logging to test_tool_with_logging and update troubleshooting documentation
- chore(conformance): remove 5 passing server tests from conformance baselines:
    - tools-call-with-logging
    - tools-call-with-progress
    - tools-call-sampling
    - tools-call-elicitation
    - elicitation-sep1034-defaults

## Motivation and Context

Two bugs in StreamableHttpServerTransport caused server-to-client notifications (e.g. notifications/message during tool calls) to be silently dropped.        
   
**Bug 1 — GET SSE stream closed immediately**                                                                                                                     
                                                                                                                                                              
`handleGetRequest` rejected GET requests with 405 when `enableJsonResponse = true` (which `mcpStreamableHttp` always sets). Since the Ktor `sse {}` handler commits response headers before the body runs, the reject failed silently and the function returned, causing Ktor to close the SSE stream. There was also no `awaitCancellation()` to keep the connection alive.                                                                                                             
                                                                                                                                                              
**Bug 2 — Notifications discarded in JSON mode**

In `send()`, notifications with a relatedRequestId entered the POST-stream path but hit `if (!isTerminated) return` without being forwarded anywhere.             
   
**Fix**                                                                                                                                                           
                                                                                                                                                              
- Remove the enableJsonResponse guard from handleGetRequest. The GET SSE stream is an always-required notification channel, orthogonal to how POST responses are delivered.                                                                                                                                              
- Add `awaitCancellation()` to keep the GET SSE connection open until the client disconnects or the transport closes.                                           
  - In `send()`, route notifications with a relatedRequestId to the standalone GET SSE stream when `enableJsonResponse = true`. 

## How Has This Been Tested?

Verified against the MCP conformance test suite — the ToolsCallWithLogging scenario now passes. ngrep confirms the `GET /mcp` stream stays open and `notifications/message` events arrive before the `tools/call` response.


## Breaking Changes
<!-- Will users need to update their code or configurations? -->

## Types of changes
<!-- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
- [ ] Documentation update

## Checklist
<!-- Go over all the following points, and put an `x` in all the boxes that apply. -->

## Additional context

Fixes #587
Requires #585 to be merged first.

### DIFF
--- a/conformance-test/README.md
+++ b/conformance-test/README.md
@@ -114,16 +114,11 @@ Tests the conformance server against all server scenarios:
 8 scenarios are expected to fail due to current SDK limitations (tracked in [
 `conformance-baseline.yml`](conformance-baseline.yml).
 
-| Scenario                              | Suite  | Root Cause                                                                                                                                             |
-|---------------------------------------|--------|--------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `tools-call-with-logging`             | server | Notifications from tool handlers have no `relatedRequestId`; transport routes them to the standalone SSE stream instead of the request-specific stream |
-| `tools-call-with-progress`            | server | *(same as above)*                                                                                                                                      |
-| `tools-call-sampling`                 | server | *(same as above)*                                                                                                                                      |
-| `tools-call-elicitation`              | server | *(same as above)*                                                                                                                                      |
-| `elicitation-sep1034-defaults`        | server | *(same as above)*                                                                                                                                      |
-| `elicitation-sep1330-enums`           | server | *(same as above)*                                                                                                                                      |
-| `resources-templates-read`            | server | SDK does not implement `addResourceTemplate()` with URI pattern matching; resources are looked up by exact URI                                         |
-| `elicitation-sep1034-client-defaults` | client | SDK does not fill in `default` values from the elicitation request schema before sending the response                                                  |
+| Scenario                              | Suite  | Root Cause                                                                                                     |
+|---------------------------------------|--------|----------------------------------------------------------------------------------------------------------------|
+| `elicitation-sep1330-enums`           | server | *(same as above)*                                                                                              |
+| `resources-templates-read`            | server | SDK does not implement `addResourceTemplate()` with URI pattern matching; resources are looked up by exact URI |
+| `elicitation-sep1034-client-defaults` | client | SDK does not fill in `default` values from the elicitation request schema before sending the response          |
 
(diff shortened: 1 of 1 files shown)


conformance-test/README.md
```diff
@@ -92,6 +92,20 @@ Tests the conformance server against all server scenarios:
...
+
+```bash
+## brew install ngrep # Install ngrep if not already installed
+
+sudo ngrep -d lo0 -W byline -s 0 port 3001  # Capture request + response bodies (SSE streams)   
+sudo ngrep -d lo0 -O capture.pcap port 3001 # Write to pcap file for later analysis   
+
+## Alternative with tcpdump if ngrep isn't installed:                                                                                                            
+sudo tcpdump -i lo0 -A -s 0 'tcp port 3001'
+```
```
devcrocod: Please don't recommend third-party tools to users if we cannot be confident that they are safe

devcrocod: Even though these tools are popular and well-known, they may still contain vulnerabilities
In my opinion, we should not recommend them in public documentation without including an appropriate disclaimer

kpavlov: okay

conformance-test/README.md
```diff
@@ -114,16 +114,11 @@ Tests the conformance server against all server scenarios:
 8 scenarios are expected to fail due to current SDK limitations (tracked in [
```
devcrocod, line 114: Please reduce the number of expected failed scenarios

kpavlov: Will do it in the next PR

Issue #591, opened 2026-03-07
Server does not support resource template parameter substitution


devcrocod:
- **Target:** server
- **Component:** sdk-server
- **Scenarios:** `resources-templates-read`

### Root cause

The SDK's `Server` / `FeatureRegistry` does not implement `addResourceTemplate()` with URI pattern matching. Resources are looked up by exact URI, so a request for `test://template/123/data` does not match the registered template `test://template/{id}/data`.

Related: #502, #503

PR #503, opened 2026-02-04
Implement resource templates

KuechA:
This PR adds supports for resource templates and is related to #502. Still work in progress, so most checks are not done yet but I'm happy to discuss feedback on the overall idea and design.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

PR #606, opened 2026-03-16
feat: #502 resource templates with basic template matcher

### LINKED ISSUE #502: Implement Support for Resource Templates
Target: server
Component: sdk-server
Scenarios: `resources-templates-read`
Specification: https://modelcontextprotocol.io/specification/2025-11-25/server/resources#resource-templates

The MCP specification discusses the possibility of using [resource templates](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#resource-templates). In my understanding, these would be useful to access/traverse resources. In my case, I'd love to use this to make a graph representation of https://github.com/Fraunhofer-AISEC/cpg available to an agent.  However, these are currently not supported by the kotlin-sdk, so I was wondering how to support them.

I would like to have resource templates implemented in the kotlin-sdk for MCP.

Alternatives are:
* Using tools instead of resources/resource templates
* Using resources and them parsing arguments in the code itself (see the suggestion [here](https://github.com/SamMorrowDrums/mcp-kotlin-starter/blob/main/SDK_LIMITATIONS.md))

**Additional context**
I tried to start with a prototypical implementation in #503 and would love to discuss alternative solutions, ideas on the design and implementation of this system and suggestions thereof.

**Current state**

The SDK's Server / FeatureRegistry does not implement addResourceTemplate() with URI pattern matching. Resources are looked up by exact URI, so a request for `test://template/123/data` does not match the registered template `test://template/{id}/data`.

kpavlov:
## Summary

> [!IMPORTANT] 
>
> **Full support of RFC-6570 is out of scope and can be implemented by users or as an independent feature.**
> **Reasoning:** Safe and secure RFC-6570 implementation is a separate problem and should not block the support of basic MCP features.

Adds resource template matching for the MCP server, enabling handlers to be registered against RFC 6570 Level 1 URI templates (e.g. `files/{path}`, `users/{id}/profile`) and resolved by specificity when multiple templates match the same URI.

- **`ResourceTemplateMatcher` / `ResourceTemplateMatcherFactory`** — extension point for URI
    template matching; `ServerOptions.resourceTemplateMatcherFactory` is public so callers can inject a custom implementation.
  - **`PathSegmentTemplateMatcher`** — default implementation: splits on `/`, parses the template once at construction, and scores each match (literal segment = 2 pts, variable capture = 1 pt) so the most-specific template always wins. Enforces safety limits (URI length cap, segment depth cap) and delegates dot-segment normalization to the platform URI parser (JVM: `java.net.URI.normalize`, JS: `URL` API, native/WASM: no-op — documented limitation).
  - **Specificity selection** — `handleReadResource` uses `maxWithOrNull` (O(n)) to select the highest-scoring match; ties broken by fewest variable captures, then registration order.
  - **fix(core): rethrow `CancellationException`** — `Protocol.kt` had a `catch (cause: Throwable)` that silently suppressed coroutine cancellation. The catch block now re-throws `CancellationException` before the generic handler runs.
  - **fix(server): `checkNotNull` in capability validation** — replaces `check(x != null)` with the idiomatic `checkNotNull(x)` across all capability guard sites in `Server.kt`.

## Security

`PathSegmentTemplateMatcher` applies one percent-decode pass via Ktor's `decodeURLPart`. Decoded variable values are attacker-controlled and documented as such — handlers must validate before using values in file paths, database queries, or downstream URLs. A dedicated security regression test suite (`PathSegmentTemplateMatcherSecurityTest`) covers double-decode, null bytes, encoded slashes, depth exhaustion, dot-segment traversal, and query/fragment pass-through.

## Motivation and Context

Closes #502, #591 

## How Has This Been Tested?

Conformance test (resources-templates-read #591), unit/integration/security test 

## Breaking Changes

Any code that called `McpException(-32600, "Connection closed")` previously received `exception.message == "MCP error -32600: Connection closed"`. After this change it returns "Connection closed". This is a behavioral API break — the `Exception.message` contract changed without a deprecation path.   

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

#603, #604, #503

### DIFF
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractPromptIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractPromptIntegrationTest.kt
@@ -9,7 +9,6 @@ import io.modelcontextprotocol.kotlin.sdk.types.GetPromptResult
 import io.modelcontextprotocol.kotlin.sdk.types.McpException
 import io.modelcontextprotocol.kotlin.sdk.types.PromptArgument
 import io.modelcontextprotocol.kotlin.sdk.types.PromptMessage
-import io.modelcontextprotocol.kotlin.sdk.types.RPCError
 import io.modelcontextprotocol.kotlin.sdk.types.Role
 import io.modelcontextprotocol.kotlin.sdk.types.ServerCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.TextContent


kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
```diff
@@ -322,19 +324,18 @@ public abstract class Protocol(@PublishedApi internal val options: ProtocolOptio
...
-                        ),
-                    ),
-                )
+                val rpcError = if (cause is McpException) {
+                    RPCError(code = cause.code, message = cause.message.orEmpty(), data = cause.data)
+                } else {
+                    RPCError(code = RPCError.ErrorCode.INTERNAL_ERROR, message = cause.message ?: "Internal error")
+                }
+                transport?.send(JSONRPCError(id = request.id, error = rpcError))
             } catch (sendError: Throwable) {
```
e5l, line 341: let's also bypass CancellationException here

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/McpException.kt
```diff
@@ -13,7 +13,7 @@ import kotlin.jvm.JvmOverloads
  */
 public class McpException @JvmOverloads public constructor(
     public val code: Int,
-    message: String,
+    message: String?,
```
e5l: Should we provide the default message instead?

Issue #572, opened 2026-03-02
StdioServer: messages should be processed concurrently


kpavlov:
Stdio transport is the most convenient option in a distributed application, since you can pass a custom sink/source to it. However, the current implementation has critical issue:

Client messages are processed sequentially. This means there can't be 2 requests being processed at once.
If there's a pinger to check client liveness, the "pong" response from the client won't be handled before the previous, maybe very long-running, request handler is finished.
Also, it blocks advanced protocol features like elicitation.

This issue was originally reported in #564, but was extracted to a separate task. cc: @i1bro

phemmer:
Sounds like #176

rnett:
This creates deadlocks for any bidirectional communication, which is now easy with `ClientConnection`.  Even for `ping`.  It's hidden in the integration tests because of `InMemoryTransport`.  `ChannelTransport` suffers from this too, FWIW.

I would propose that the `Protocol` forks (i.e. `launch`s) the request handlers rather than relying on transport-specific handling.

rnett:
@kpavlov do you plan on addressing this soon?  It's blocking my upgrade.  If not, would you accept a simple PR to use launch in Protocol?

kpavlov:
@rnett , please create the PR. It might take some time to redesign things properly. Now we are focusing on full protocol features support + conformance tests. Once we have test coverage- it will be much easier to refactor.

rnett:
PR: https://github.com/modelcontextprotocol/kotlin-sdk/pull/610.  Didn't end up being able to use `Protocol` for reasons explained there.

rnett:
There are some issues with the PR, and I have not had a chance to work on it (and may not for some time).  Feel free to take it over or close it if you fix it yourselves.

kpavlov:
Thank you, @rnett 
This is my priority now

PR #610, opened 2026-03-18
fix: concurrent message processing for all transports

rnett:
## Motivation and Context

Adds somewhat-unified concurrent request handling as mentioned in https://github.com/modelcontextprotocol/kotlin-sdk/issues/572 and related issues.  Unfortunately it wasn't possible to handle this in `Protocol`, since some transports (sse server mostly) need to handle the request inline and are already in a pre-request coroutine.  I went with the next best place I could find, which was `AbstractTransport`.

I do think there's room for improvement here, including:
 * More centralized handling of this sort of thing, the coroutine scopes, and error handling. I would think the best place is in `Protocol` but the current lateinit setup architecture doesn't allow for it.
 * Unified graceful shutdown with timeout + cancellation
 * Updating `setNotificationHandler` and friends to take a suspending lambda, not one that returns `Deferred`.

However, as per the discussion on the issue, I wanted to keep this change as minimal as possible.

## How Has This Been Tested?
All unit and integration tests pass, and I added tests to ensure that suspending and blocking handlers would not block other calls.
`testToolConcurrentProcessing` in particular ensures that new messages can be handled while tool execution is ongoing.  It fails for stdio and channel on `main`.

## Breaking Changes
None.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -77,6 +77,9 @@ Follow these rules to keep changes safe, comprehensible, and easy to maintain.
 - **Prioritize test readability**
 - Avoid creating too many test methods; use parametrized tests when testing multiple similar scenarios
 - When running tests on Kotlin Multiplatform projects, run JVM tests only unless asked for other platforms
+- **Concurrency in Tests**: Always use thread-safe collections (e.g., `Mutex`-protected lists or `Channel`) when
+  collecting messages from transports that process messages concurrently in the background (like those inheriting from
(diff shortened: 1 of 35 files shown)


kpavlov:
@rnett, please review #670. Does it improve the situation?

rnett:
Yeah, I think that would solve it.

Issue #437, opened 2025-11-28
SseClientTransport isn't closed when underlying sseSession is disconnected


nerzhulart:
This makes impossible to realize when a server is stopped to shutdown the client.
E.g. in Idea we can't shutdown stdio wrapper properly when Idea itself (with MCP server) is closed.

The naive solution is something like
```
Index: kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt	(revision 29d89abf0290b40f3f4e774e1617564f0e13e75b)
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SSEClientTransport.kt	(date 1764369820391)
@@ -3,6 +3,7 @@
 import io.github.oshai.kotlinlogging.KotlinLogging
 import io.ktor.client.HttpClient
 import io.ktor.client.plugins.sse.ClientSSESession
+import io.ktor.client.plugins.sse.SSE
 import io.ktor.client.plugins.sse.sseSession
 import io.ktor.client.request.HttpRequestBuilder
 import io.ktor.client.request.post
@@ -16,6 +17,7 @@
 import io.ktor.http.protocolWithAuthority
 import io.modelcontextprotocol.kotlin.sdk.shared.AbstractTransport
 import io.modelcontextprotocol.kotlin.sdk.shared.TransportSendOptions
+import io.modelcontextprotocol.kotlin.sdk.types.Implementation
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCMessage
 import io.modelcontextprotocol.kotlin.sdk.types.McpJson
 import kotlinx.coroutines.CancellationException
@@ -28,6 +30,7 @@
 import kotlinx.coroutines.cancel
[...]

PR #612, opened 2026-03-18
docs: update AGENTS.md

devcrocod:
Improve AGENTS.md

## Motivation and Context
The previous AGENTS.md contained many elements contradicting [best practices](https://code.claude.com/docs/en/best-practices).

AGENTS.md/CLAUDE.md is loaded into the context on every request, so a ~300+ line file is too large and fills the context window.
> Keep it concise. For each line, ask: “Would removing this cause Claude to make mistakes?” If not, cut it. Bloated CLAUDE.md files cause Claude to ignore your actual instructions!

> Most best practices are based on one constraint: Claude’s context window fills up fast, and performance degrades as it fills.

> The over-specified CLAUDE.md. If your CLAUDE.md is too long, Claude ignores half of it because important rules get lost in the noise.

Since the file is loaded on every request, it should be general-purpose rather than targeting specific use cases.

> CLAUDE.md is loaded every session, so only include things that apply broadly.

 AI coding agents are used not only for writing tests and implementations.

- AGENTS.md should not contain behavioral instructions, especially like `Tests first, always.`. This causes the agent to trigger test runs even when not required, increasing request time and token consumption significantly due to large test outputs. For example, on a simple codebase question, the agent ran the full test suite and I waited over 10 minutes for a response.
- `Uphold SOLID principles` — self-evident rules, explaining what SOLID stands for is redundant since AI coding agents are already trained on this knowledge.
> Exclude: "Standard language conventions Claude already knows"

> Exclude: "Self-evident practices like 'write clean code'"

- Mentioning specific MCP servers (jetbrains, intellij-index) can mislead the agent if they are not installed in the environment. Additionally, [Anthropic recommends](https://resources.anthropic.com/hubfs/The-Complete-Guide-to-Building-Skill-for-Claude.pdf) mentioning the use of MCP in skills.
- The large Kotest patterns block targets a specific use case and is unnecessary for other agent interactions. Moreover, the agent already looks at neighboring test classes and can learn patterns from existing tests
- The large Module documentation block covers a very rare use case.
- The KDoc block — the agent can derive this information from the codebase, this should be a skill instead

These and other changes follow the best practices.

## Types of changes
- [x] Documentation update

## Checklist

## Additional context
https://code.claude.com/docs/en/best-practices
https://resources.anthropic.com/hubfs/The-Complete-Guide-to-Building-Skill-for-Claude.pdf
https://agents.md/#examples

### DIFF
--- a/conformance-test/conformance-baseline.yml
+++ b/conformance-test/conformance-baseline.yml
@@ -1,7 +1,6 @@
 # Conformance test baseline - expected failures
 # Add entries here as tests are identified as known SDK limitations
-server:
-  - resources-templates-read
+server: []
 
 client:
   - elicitation-sep1034-client-defaults
--- a/conformance-test/src/main/kotlin/io/modelcontextprotocol/kotlin/sdk/conformance/ConformanceResources.kt
+++ b/conformance-test/src/main/kotlin/io/modelcontextprotocol/kotlin/sdk/conformance/ConformanceResources.kt
@@ -45,18 +45,17 @@ fun Server.registerConformanceResources() {
     }
 
     // 3. Template resource
-    // Note: The SDK does not currently support addResourceTemplate().
-    // Register as a static resource; template listing is handled separately.
-    addResource(
-        uri = "test://template/{id}/data",
+    addResourceTemplate(
+        uriTemplate = "test://template/{id}/data",
         name = "template",
         description = "A template resource for testing",
         mimeType = "application/json",
-    ) { request ->
+    ) { request, variables ->
+        val id = variables["id"]
         ReadResourceResult(
             listOf(
                 TextResourceContents(
-                    text = "content for ${request.uri}",
+                    text = """{"id": "$id"}""",
                     uri = request.uri,
                     mimeType = "application/json",
                 ),
--- a/integration-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ClientTest.kt
+++ b/integration-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ClientTest.kt
@@ -1,5 +1,6 @@
 package io.modelcontextprotocol.kotlin.sdk.client
 
+import io.kotest.matchers.shouldBe
 import io.modelcontextprotocol.kotlin.sdk.server.Server
 import io.modelcontextprotocol.kotlin.sdk.server.ServerOptions
 import io.modelcontextprotocol.kotlin.sdk.server.ServerSession
@@ -266,8 +267,8 @@ class ClientTest {
             client.connect(failingTransport)
         }
 
-        assertEquals(-32600, exception.code)
-        assertEquals("MCP error -32600: Invalid Request", exception.message)
+        exception.code shouldBe -32600
+        exception.message shouldBe "Invalid Request"
 
         assertTrue(closed)
     }
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractPromptIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractPromptIntegrationTest.kt
@@ -1,6 +1,7 @@
 package io.modelcontextprotocol.kotlin.sdk.integration.kotlin
 
 import io.kotest.assertions.withClue
+import io.kotest.matchers.shouldBe
 import io.kotest.matchers.string.shouldContain
 import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequest
 import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequestParams
@@ -8,7 +9,6 @@ import io.modelcontextprotocol.kotlin.sdk.types.GetPromptResult
 import io.modelcontextprotocol.kotlin.sdk.types.McpException
 import io.modelcontextprotocol.kotlin.sdk.types.PromptArgument
 import io.modelcontextprotocol.kotlin.sdk.types.PromptMessage
-import io.modelcontextprotocol.kotlin.sdk.types.RPCError
 import io.modelcontextprotocol.kotlin.sdk.types.Role
 import io.modelcontextprotocol.kotlin.sdk.types.ServerCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.TextContent
@@ -688,13 +688,13 @@ abstract class AbstractPromptIntegrationTest : KotlinTestBase() {
             }
         }
 
-        val expectedMessage = "MCP error -32603: Prompt not found: non-existent-prompt"
+        val expectedMessage = "Prompt not found: non-existent-prompt"
 
-        assertEquals(
-            RPCError.ErrorCode.INTERNAL_ERROR,
-            exception.code,
-            "Exception code should be INTERNAL_ERROR: ${RPCError.ErrorCode.INTERNAL_ERROR}",
-        )
-        assertEquals(expectedMessage, exception.message, "Unexpected error message for non-existent prompt")
+        withClue("Exception code should be INTERNAL_ERROR: -32603") {
+            exception.code shouldBe -32603
+        }
+        withClue("Unexpected error message for non-existent prompt") {
+            exception.message shouldBe expectedMessage
+        }
     }
 }
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractResourceIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/kotlin/AbstractResourceIntegrationTest.kt
@@ -213,14 +213,11 @@ abstract class AbstractResourceIntegrationTest : KotlinTestBase() {
             }
         }
 
-        val expectedMessage = "MCP error -32603: Resource not found: test://nonexistent.txt"
-
         assertEquals(
-            RPCError.ErrorCode.INTERNAL_ERROR,
+            RPCError.ErrorCode.RESOURCE_NOT_FOUND,
             exception.code,
-            "Exception code should be INTERNAL_ERROR: ${RPCError.ErrorCode.INTERNAL_ERROR}",
+            "Exception code should be RESOURCE_NOT_FOUND: ${RPCError.ErrorCode.RESOURCE_NOT_FOUND}",
         )
-        assertEquals(expectedMessage, exception.message, "Unexpected error message for invalid resource URI")
     }
 
     @Test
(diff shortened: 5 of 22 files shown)


kpavlov (changes requested):
I agree, that project information was a bit redundant. Everything else is a degradation.
These instructions are now useless and will negatively affect code quality. Let's revert requirements, they are useful and should be loaded on every conversion.

devcrocod:
@kpavlov 
I included a number of arguments in the PR description, have you had a chance to review them?
I did not go through every single line of the previous agents.md, because that would have taken even more time. However, I can justify every removal by referring to the relevant recommendations.

Could you please explain in more detail why you are blocking this PR?

CLAUDE.md
kpavlov, line 1: Symlinks may not work on all file systems. Users might need to configure if explicitly with: `git config --global core.symlinks true`. `@AGENTS.md` is include instruction, working everywhere.

kpavlov:
@devcrocod, yes, I reviewed your comments and understand that the 10-minute wait time is definitely an issue. But the AGENTS.md directly shapes every line of code the project produces. A wrong removal is not a one-time mistake — it removes a guardrail from every future contribution. That's why each instruction deserves individual justification against observed agent behavior, not just against general recommendations.                        

Regarding the change: too many important instructions were removed on the way.
No skills were added, and I'm not sure whether other agents understand skills as well as Claude does. If your intent is to move something to skills, let's add the skills first.

- **Let's re-add "always add tests for new features and bug fixes, even when not explicitly asked".** This is not about "tests first" (which caused the 10-min wait problem and was correctly removed). "always add tests even when not explicitly asked" - is a non-default behavior. Without it, the agent will write the implementation and stop.

- **Add a pointer to `verifySerialization` / `verifyDeserialization`** - these are project-specific helpers. The agent cannot discover them without reading all the `test-utils` source. 

- **Using tools is not obvious, but it could be a skill.** I observed Claude not using IDE tools but doing grep instead and getting inaccurate results, that's why this instruction was added. I don't think this instruction would harm when such tools are not available?

- **Avoid creating too many test methods; use parametrized tests when testing multiple similar scenarios** - this is not-default behavior

- "Kotest patterns block (shouldBe, shouldNotBeNull {}, shouldMatchJson)" - The style is non-default and non-obvious. Projects have different test styles. This should stay. 

- Let's add instructions about adhering to Kotlin structural concurrency, something like "Follow Kotlin structured concurrency: every coroutine has a defined scope and lifetime, CancellationException is always re-thrown, and Mutex is never held across suspension points."

I'm not objecting to removing most of the format requirements. But guardrails about test coverage must stay.

devcrocod:
@kpavlov 

> That's why each instruction deserves individual justification against observed agent behavior, not just against general recommendations.

This assumes that the previous agent.md was written correctly, thoughtfully, and in line with best practices from the actual developers of coding agents. I strongly doubt that was the case.

First, I highly doubt the previous instructions were meaningfully justified at all.
Second, as I explained in the PR description, many parts of the old file: such as hard directives, explanations of SOLID principles to a coding agent, large blocks for rare edge cases, and heavy token consumption for unrelated tasks — suggest that these were not isolated mistakes. They indicate that the previous agents.md was written without a real understanding of how coding agents actually work.
Third, these "general recommendations" are not just random articles from the internet, they come directly from the creators of coding agents.

> Regarding the change: too many important instructions were removed on the way.

I would like to understand on what basis they are considered important. Unfortunately, I do not see any arguments supporting that view. On the contrary, Anthropic's recommendations, the examples in agents.md, and a direct review of the previous version of the file with the coding agent seem to suggest the opposite.

> No skills were added, and I’m not sure whether other agents understand skills as well as Claude does. If your intent is to move something to skills, let’s add the skills first.

This PR addresses a concrete problem: a poor-quality agents.md. I do not think the lack of skills should block this change, because the previous agents.md is actively harmful for contributors. It wastes time even on simple requests, produces worse results by filling the context window with contradictory and low-value instructions, and, importantly, consumes a large number of tokens, which directly costs money. I already included specific quotes from Anthropic’s recommendations on this point.

Now, regarding the specific instructions you mentioned:

> Let's re-add "always add tests for new features and bug fixes, even when not explicitly asked"

No, this is not behavior that differs from the default. You can look at how coding models are trained. The testing section in agents.md also makes it clear that writing tests is necessary. Also, this is again a behavioral instruction, and I would refer you back to the PR description and Anthropic’s recommendations.

A simple example: implementation can be iterative, and test creation may be best left to the final request, once the coding agent has the full implementation context. Forcing test creation after every request is redundant and can even reduce test quality, because the agent may not yet have the full picture of the final implementation.

> Add a pointer to verifySerialization / verifyDeserialization

You said: "The agent cannot discover them without reading all the test-utils source."

This is easy to verify. I ran the agent, and it found that these utilities are defined in test-utils and used in 16 test files. They are easily discoverable — the agent can see `import io.modelcontextprotocol.kotlin.test.utils.verifySerialization` in any of those nearby test files.

> Using tools (MCP servers jetbrains/intellij-index)

These are external dependencies. Mentioning external dependencies in agents.md, which is appended to every request, is harmful. The agent starts searching for those servers directly, which wastes both time and context window space. 

Also, I would not recommend using the jetbrains mcp server without precise tool-usage instructions in skills. You can review the available tools in those mcp servers and also this Anthropic article, in addition to the ones I already linked: https://www.anthropic.com/engineering/code-execution-with-mcp

> Avoid creating too many test methods; use parametrized tests when testing multiple similar scenarios

Again, I would refer you to the PR description. This is not a universal instruction. Also, a coding agent already reads nearby test files to understand test structure, style, and conventions, so it will pick this up naturally from the codebase when relevant.

> Kotest patterns block (shouldBe, shouldNotBeNull {}, shouldMatchJson)

The same argument applies here as above. In addition, the testing frameworks are explicitly listed there, and the infix functions already make the intended usage pattern fairly obvious.

> Let’s add instructions about adhering to Kotlin structural concurrency

A quick search across the codebase did not show obvious violations of this. Were there actual problems caused by agents violating this rule? Do we have concrete examples of such incidents?

kpavlov:
> This assumes that the previous agent.md was written correctly, thoughtfully, and in line with best practices from the actual developers of coding agents. I strongly doubt that was the case.

This is a wrong assumption, agent.md was not perfect. But this PR removes almost everything.

Regarding tests: in practice, agents frequently stop after writing the implementation without adding tests, even for bug fixes. The current testing section describes how to write and run tests (frameworks, patterns), but doesn't instruct the agent to write them unprompted. That's the gap.

>> Let’s add instructions about adhering to Kotlin structural concurrency
>
> A quick search across the codebase did not show obvious violations of this. Were there actual problems caused by agents violating this rule? Do we have concrete examples of such incidents?

We have a bunch of open issues (#242, #564, #437), and some (missed) CancellationException handling already fixed. This was somehow missed by the agent.

--
I'm happy to approve if we:                                               

1. Add one line on test coverage: "Always add tests for new features and bug fixes, even when not explicitly asked."                                 

2. Add one line on structured concurrency: "Follow Kotlin structured  concurrency: every coroutine has a defined scope and lifetime, CancellationException is always re-thrown, and Mutex is never held across suspension points."                                                         
                            
Everything else in your PR I can accept. These two points address observed agent failures, not theoretical ones.

AGENTS.md
```diff
@@ -1,334 +1,80 @@
...
-    enum shouldHaveSize 2 // Preferred
-    this.enum shouldHaveSize 2 // Avoid
-}
+```bash
+./gradlew ktlintCheck                # Lint check
+./gradlew ktlintFormat               # Auto-fix lint issues
+./gradlew apiCheck                   # Check public API compatibility (run before committing)
+./gradlew apiDump                    # Update API dump after intentional API changes
+./gradlew koverLog                   # Print coverage summary
+./gradlew koverHtmlReport            # Generate HTML coverage report
```
kpavlov:
```suggestion
./gradlew koverXmlReport            # Generate HTML coverage report
```
Xml a single machine readable file

AGENTS.md
```diff
@@ -1,334 +1,80 @@
...
-#### Avoiding `this.` in Lambda Blocks
-
-```kotlin
-// Avoid explicit `this.` unless necessary for disambiguation
-prop.shouldNotBeNull {
-    enum shouldHaveSize 2 // Preferred
-    this.enum shouldHaveSize 2 // Avoid
-}
+```bash
+./gradlew ktlintCheck                # Lint check
```
kpavlov, line 12:
```suggestion
./gradlew detekt ktlintCheck                # Lint check
```

CLAUDE.md
e5l: Referenced files are working in Claude

e5l:
We can also keep cancellation exception point, the agent is not doing this well

AGENTS.md
devcrocod: yeap, okay

AGENTS.md
devcrocod: 👍 
Okay, I think I'll put this on a separate line so it doesn't get mixed up

devcrocod:
> This is a wrong assumption, agent.md was not perfect. But this PR removes almost everything.

That was a direct response to the previous comment:

> A wrong removal is not a one-time mistake — it removes a guardrail from every future contribution. That's why each instruction deserves individual justification against observed agent behavior, not just against general recommendations.

> But this PR removes almost everything.

Exactly. And I believe I provided sufficient reasoning for why that is justified.

> Regarding tests: in practice, agents frequently stop after writing the implementation without adding tests, even for bug fixes. The current testing section describes how to write and run tests (frameworks, patterns), but doesn't instruct the agent to write them unprompted. That's the gap.

I already addressed this earlier, but I'll repeat it once more: behavioral instructions should not be added to the general agents.md. Please refer to my previous comments. Also, I already gave an example where requiring tests after every request can actually make generation worse rather than better. Other examples include conformance tests and test utils. A behavioral instruction like this would create false positives in those cases.

> We have a bunch of open issues (https://github.com/modelcontextprotocol/kotlin-sdk/issues/242, https://github.com/modelcontextprotocol/kotlin-sdk/issues/564, https://github.com/modelcontextprotocol/kotlin-sdk/issues/437), and some (missed) CancellationException handling already fixed. This was somehow missed by the agent.

Yes, that is indeed worth adding

CLAUDE.md
devcrocod: It works
Please take a look here: https://github.com/modelcontextprotocol/modelcontextprotocol

AGENTS.md
kpavlov: That's fine, but IMO running all checks in one short would be cheaper. We can keep it

devcrocod:
@kpavlov 
> Let's add that the agent should ensure the added & modified code is properly tested.

I’m going to merge this PR as it is for now, so we don't delay the process.
After that, you can open a new PR with this addition.
In my view, it would be better to present this not as a behavioral instruction, but to rephrase it in a more descriptive way. As I mentioned above, behavioral instructions do not always work reliably and may conflict with contributors’ pipelines.
For example, if their testing process is based on TDD or, conversely, is left until the very end before commit

PR #617, opened 2026-03-25
chore: simplify and clean up CI workflows 

devcrocod:
- Remove redundant Gradle flags, env vars, and unused settings across all workflows
- Minimize GitHub Actions permissions to least-privilege
- Drop Windows build matrix (JVM tests covered by Linux)
- Simplify CodeQL workflow (remove matrix, unused permissions, stale triggers)
- Clean up gradle.properties (remove unused KMP settings)

## Breaking Changes
—

## Types of changes
- [ ] Documentation update

## Checklist


.github/workflows/apidocs.yaml
```diff
@@ -25,9 +25,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
-        with:
```
kpavlov, line 28: 👍🏻 thank you

.github/workflows/build.yml
```diff
@@ -127,7 +124,7 @@ jobs:
       - name: Archive Maven Local Repository
         if: ${{ success() && matrix.os == 'ubuntu-latest' }}
         run: |
-          cd ~/.m2/repository && zip -r $GITHUB_WORKSPACE/local-maven-repository.zip .
+          cd ~/.m2/repository/io/modelcontextprotocol && zip -r $GITHUB_WORKSPACE/local-maven-repository.zip .
```
kpavlov, line 127: This will cause other artifacts to be re-downloaded in later stages.

.github/workflows/codeql.yml
```diff
@@ -1,59 +1,42 @@
 name: "CodeQL Advanced"
 
 on:
-  push:
-    branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
   schedule:
-    - cron: '0 4 * * 0'
+    - cron: '0 7 * * 1' # Monday 8:00 CET (7:00 UTC)
```
kpavlov, line 7: Why not daily? Is it expensive to run?

gradle.properties
```diff
@@ -1,16 +1,10 @@
...
 org.gradle.parallel=true
 org.gradle.caching=true
-# Dokka
-org.jetbrains.dokka.experimental.gradle.pluginMode=V2Enabled
-org.jetbrains.dokka.experimental.gradle.pluginMode.noWarn=true
 # Kotlin
 kotlin.code.style=official
 kotlin.daemon.jvmargs=-Xmx4G
-# Build JS targets using npm package manager https://kotlinlang.org/docs/js-project-setup.html#npm-dependencies
-kotlin.js.yarn=false
```
kpavlov, line 11: Let us revert this as yarn is an optional dependency and may not be installed.

devcrocod: On the contrary, as the documentation states:
 
> By default, to download and install your declared dependencies at build time, the plugin manages its own instance of the [Yarn](https://yarnpkg.com/lang/en/) package manager. It works out of the box without additional configuration, but you can tune it or use Yarn already installed on your host.

https://kotlinlang.org/docs/js-project-setup.html#yarn

In other words, this option disables the built-in automatic use of Yarn. As a result, users are required to install npm manually

.github/workflows/codeql.yml
devcrocod: It was already weekly by default. I only moved it from Sunday to Monday.

There’s no need to run it daily, that would be unnecessary

.github/workflows/build.yml
devcrocod: Yes, and only the ones that are actually needed

1. In my view, that also includes checking the dependencies that are actually used in samples
2. The root project brings in many more dependencies that the samples do not need
3. The samples do not have that many dependencies, so downloading them again is not an expensive operation

PR #658, opened 2026-03-30
feat(docs): public api kdocs

devcrocod:
- enable rule for checking undocumented public api 
- add kdoc skill

## How Has This Been Tested?
./gradlew detekt

## Breaking Changes
none

## Types of changes
- [x] Documentation update

## Checklist

### DIFF
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnectionTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnectionTest.kt
@@ -7,6 +7,8 @@ import io.kotest.matchers.string.shouldNotBeBlank
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequest
 import io.modelcontextprotocol.kotlin.sdk.types.CallToolRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.ClientCapabilities
+import io.modelcontextprotocol.kotlin.sdk.types.ElicitationCompleteNotification
+import io.modelcontextprotocol.kotlin.sdk.types.ElicitationCompleteNotificationParams
 import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequest
 import io.modelcontextprotocol.kotlin.sdk.types.GetPromptRequestParams
 import io.modelcontextprotocol.kotlin.sdk.types.ListRootsRequest
@@ -77,6 +79,8 @@ class ClientConnectionTest : AbstractServerFeaturesTest() {
         val loggingMessage = onClientNotification<LoggingMessageNotification>(Method.Defined.NotificationsMessage)
         val resourceUpdated =
             onClientNotification<ResourceUpdatedNotification>(Method.Defined.NotificationsResourcesUpdated)
+        val elicitationComplete =
+            onClientNotification<ElicitationCompleteNotification>(Method.Defined.NotificationsElicitationComplete)
 
         init {
             onClientRequest<ListRootsRequest, ListRootsResult>(Method.Defined.RootsList) {
@@ -109,6 +113,12 @@ class ClientConnectionTest : AbstractServerFeaturesTest() {
                 withClue("sendResourceUpdated() delivered wrong URI") {
                     resourceUri shouldBe "test://res"
                 }
+                withClue("sendElicitationComplete() did not deliver a notification to the client") {
+                    elicitationComplete.isCompleted shouldBe true
+                }
+                withClue("sendElicitationComplete() delivered wrong elicitationId") {
+                    elicitationComplete.await().params.elicitationId shouldBe "elicit-123"
+                }
                 capturedSessionId.shouldNotBeBlank()
             }
         }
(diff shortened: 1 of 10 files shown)


.claude/skills/kdoc/SKILL.md
```diff
@@ -0,0 +1,74 @@
+---
+name: kdoc
+description: "Add KDoc documentation to Kotlin public API. Use whenever the user asks to document Kotlin code, add KDoc, generate API docs, mentions undocumented public declarations, or wants to improve existing documentation. Also trigger when the user says 'add docs', 'document this class/file/module', 'write KDoc', or asks about missing documentation in Kotlin code."
```
kpavlov, line 3: This should also trigger automatically when adding or updating public facing api. User will simply forget to ask about kdoc.

.claude/skills/kdoc/SKILL.md
```diff
@@ -0,0 +1,74 @@
...
+Add KDoc comments to public Kotlin API declarations.
+
+## What to document
+
+All public declarations (no explicit `private`/`internal`/`protected`):
+- Classes, interfaces, objects, sealed classes/interfaces
+- Functions, extension functions
+- Properties
+- Enum classes and entries
+- Type aliases, annotation classes
```
kpavlov, line 17: It this looks like too obvious

.claude/skills/kdoc/SKILL.md
```diff
@@ -0,0 +1,74 @@
...
+## KDoc format
+
+**Class/interface example:**
+
+````kotlin
+/**
+ * Manages active client sessions and their lifecycle.
+ *
+ * Sessions are created on first connection and cleaned up
+ * when the transport closes.
```
kpavlov, line 34: Adding example for a class also make sense, e.g. how to construct the class

.claude/skills/kdoc/SKILL.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Rules
+
+- Summary: concise first sentence starting with a third-person verb ("Creates", "Returns", "Represents"). Expand to 2-3 sentences only when genuinely complex
+- **@property** in class-level KDoc for all public properties (never as individual KDoc on the property); **@param** for function parameters
+- **@return** for non-Unit return types
+- **Example block**: add for DSL builders, complex functions, extension functions with non-obvious usage. Skip for trivial one-liners and simple getters
+- **DSL builders**: always include an Example showing full usage with the receiver scope — this is critical for discoverability
+- KDoc links (`[ClassName]`, `[methodName]`): only where it adds clear navigational value
+- No **@throws** — don't document exceptions
```
kpavlov, line 72: This is questionable. It is very helpful to keep exceptions documented, especially for non-trival use cases

config/detekt/detekt.yml
```diff
@@ -22,3 +22,16 @@ empty-blocks:
   EmptyFunctionBlock:
     excludes: *testFolders
 
+comments:
+    UndocumentedPublicClass:
+      active: true
+      excludes: [ '**/test/**', '**/commonTest/**', '**/jvmTest/**', '**/generated-sources/**' ]
```
kpavlov: Consider referring `*testForders`

.claude/skills/kdoc/SKILL.md
devcrocod: - this violates the solid principles, a skill should have a single responsibility - generating kdoc
- Automatic invocation triggers do not work that way. You can create a separate orchestration skill for your own use or add a dedicated hook. However, keep in mind that a hook will introduce side effects unless you build a proper workflow around it
- That is exactly what detekt rules are for, to make sure the user does not forget about this and that such a pr does not get merged

.claude/skills/kdoc/SKILL.md
devcrocod: This is not obvious to the agent. In the test eval, the agent missed some points, so I add it explicitly

.claude/skills/kdoc/SKILL.md
devcrocod: It seems that constructor signatures are enough, at least for now, while we are using simple constructors

.claude/skills/kdoc/SKILL.md
devcrocod: Agreed. Exceptions should be documented in non-trivial cases or when custom exceptions are involved. In kotlin-sdk, there is only `McpException`. it is thrown by two methods, and this is already documented there

.claude/skills/kdoc/SKILL.md
kpavlov: This doesn't violate solid principles, only makes the skill-use condition more explicit. Detekt definitely helps, but it may consume more tokens on the first failing gradle run than this skill consumes.

Did you verify that if you ask the Agent to write a new method or to update the public method without KDoc, the skill is applied and the required KDoc is generated?

Not blocking for this PR but a room for improvement

kpavlov (dismissed):
Let's ship it!
This PR adds missing KDocs (even if it itself violates SOLID principles for PRs). Skill is helpful but might the triggering mechanism might need improvement or not.

kpavlov (approved):

PR #665, opened 2026-04-01
feat(server): auto-install ContentNegotiation with McpJson (#664)

### LINKED ISSUE #664: Server silently accepts wrong ContentNegotiation JSON configuration, causing runtime serialization failures
## Description

Affected version: 0.10.0

When users set up a Ktor MCP server, the README and API surface require them to manually install ContentNegotiation with the correct McpJson configuration:                                                      
    
```kotlin  
embeddedServer(...) {                                                                                                                                                                                                             
  install(ContentNegotiation) {                                                                                                                                                                                    
      json(McpJson)                                                                                            
  }  
  mcpStreamableHttp {
      ...
  }
}                                                                                                                                                                                                              
```   

If a user installs ContentNegotiation with the default `json()` (no `McpJson`), or omits `ContentNegotiation` entirely, the server silently accepts the misconfiguration. Failures surface only at runtime as subtle JSON-RPC serialization errors:                                                                                                                                                                                 
                                                                                                                                                                                                                   
  - explicitNulls = true (the default) emits "field": null in responses, violating MCP's requirement that null fields are absent                                                                                   
  - encodeDefaults = false (the default) omits fields that MCP clients expect to always be present
  - classDiscriminatorMode mismatches cause type-discriminator pollution in polymorphic payloads                                                                                                                   
                                                                                                                                                                                                                   
These errors are difficult to diagnose because the server starts successfully and many requests appear to work. Failures only occur on specific message shapes.                                                 
                                                                                                                                                                                                                   
## Root Cause                                                                                                                                                                                                       
                                                                                                                                                                                                                 
`mcpStreamableHttp()` and `mcpStatelessStreamableHttp()` in `KtorServer.kt` delegate JSON configuration entirely to the user. There is no validation, auto-configuration, or warning when the wrong serialization settings are active.

## Proposed Fix                                                                                                                                                                                                   

  1. Auto-install `ContentNegotiation` with `McpJson` inside `mcpStreamableHttp()` / `mcpStatelessStreamableHttp()`, so the correct config is guaranteed without user action                                               
  2. Warn at startup if `ContentNegotiation` was already installed by user code                                                                                                                                                                                   
  3. Update README to remove the manual install(ContentNegotiation) instructions and document that the SDK handles it automatically

kpavlov:
Users currently must manually `install(ContentNegotiation) { json(McpJson) }` before calling `mcp()`,
`mcpStreamableHttp()`, or `mcpStatelessStreamableHttp()`. Forgetting this or using the default `json()`    
instead of `json(McpJson)` causes silent serialization failures at runtime (e.g., explicit null fields in JSON-RPC
responses). This PR makes the SDK install it automatically.

Fixes #664

## Changes

- **Core feature** — `mcp()`, `mcpStreamableHttp()`, and `mcpStatelessStreamableHttp()` now auto-install `ContentNegotiation` configured with `McpJson`. If the user already installed `ContentNegotiation`, a warning is logged instead of crashing with  `DuplicatePluginException`.

- **`KtorServerHelpers.kt`** (new) — `installMcpContentNegotiation()` internal helper with `pluginOrNull` guard and warning
- **`KtorServer.kt`** — calls the helper from all three entry points
- **`build.gradle.kts`** — adds `ktor-server-content-negotiation` and `ktor-serialization` as `implementation` deps to `kotlin-sdk-server`

- **Removed manual installs** from `ConformanceServer.kt` and `AbstractStreamableHttpIntegrationTest.kt`.

- **Test infrastructure** — replaced `slf4j-simple` with Logback and use `ListAppender` (`LogCapture` utility) for deterministic log assertion. Added `logback-test.xml` matching previous log levels.

- **`StreamableHttpServerConfigurationTest.kt`** (new) — verifies the warning for all three entry points: `mcpStreamableHttp`, `mcpStatelessStreamableHttp`, and `mcp` (SSE).

- **README.md** — removed manual `ContentNegotiation` boilerplate from examples, documented automatic installation, added CORS configuration snippet for browser-based clients.

- **CI** — added `ktlintCheck` to the Linux build matrix job.

## How Has This Been Tested?

Unit/integration tests. Conformance tests updated

## Breaking Changes

No

## Types of changes
- [x] Documentation update

## Checklist


.github/workflows/build.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         include:
           - os: ubuntu-latest
             job-name: "Linux Tests"
-            gradle-tasks: "build"
+            gradle-tasks: "ktlintCheck build"
```
devcrocod, line 37: How is it related to PR?
and `ktlintCheck` runs with `check`, which means it's already included in build task

gradle/libs.versions.toml
```diff
@@ -67,6 +68,7 @@ mockk = { module = "io.mockk:mockk", version.ref = "mockk" }
 mokksy = { group = "dev.mokksy", name = "mokksy", version.ref = "mokksy" }
 netty-bom = { group = "io.netty", name = "netty-bom", version.ref = "netty" }
 slf4j-simple = { group = "org.slf4j", name = "slf4j-simple", version.ref = "slf4j" }
+logback-classic = { group = "ch.qos.logback", name = "logback-classic", version.ref = "logback" }
```
devcrocod, line 71: Why do we need another logger?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServerHelpers.kt
```diff
@@ -0,0 +1,38 @@
...
+ */
+internal fun Application.installMcpContentNegotiation() {
+    if (pluginOrNull(ContentNegotiation) != null) {
+        // ContentNegotiation was already installed by user code.
+        logger.warn {
+            "ContentNegotiation plugin is already installed with an unknown JSON configuration. " +
+                "MCP requires json(McpJson) for correct JSON-RPC serialization " +
+                "(explicitNulls = false, encodeDefaults = true). " +
+                "Remove your install(ContentNegotiation) { … } block and let " +
+                "mcp() / mcpStreamableHttp() / mcpStatelessStreamableHttp() configure it automatically."
```
devcrocod: To be honest, that’s a strange suggestion for me

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -86,11 +86,15 @@ public fun Route.mcp(block: ServerSSESession.() -> Server) {
  * over [Server-Sent Events (SSE) Transport](https://modelcontextprotocol.io/specification/2024-11-05/basic/transports#http-with-sse)
  * and sets up routing with the provided configuration block.
  *
+ * Automatically installs [ContentNegotiation][io.ktor.server.plugins.contentnegotiation.ContentNegotiation]
+ * with [McpJson][io.modelcontextprotocol.kotlin.sdk.types.McpJson] and [SSE].
```
devcrocod, line 90: Why hasn’t kdoc been updated for the other methods?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
```diff
@@ -103,6 +107,7 @@ private fun Application.mcpStreamableHttp(
     configuration: StreamableHttpServerTransport.Configuration,
     block: RoutingContext.() -> Server,
 ) {
+    installMcpContentNegotiation()
```
devcrocod, line 110: If the user is configured to use multiple extensions, they will see a warning saying that ContentNegotiation is already installed

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/LogCapture.kt
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    val messages: List<String>
+        get() = appender.list.map { it.formattedMessage }
+
+    override fun close() {
+        logger.detachAppender(appender)
+        appender.stop()
+    }
+}
```
devcrocod, line 35: Wouldn't it be enough to just verify that no DuplicatePluginException is thrown?
The existing integration tests already prove that this works, since they did not break after the manual installation of ContentNegotiation was removed

Verifying the exact log output does not seem quite right because it may change at any moment. On top of that, it adds another dependency, extra configuration, and new tests that verify log output. It feels unnecessary

README.md
```diff
@@ -809,11 +806,30 @@ embeddedServer(CIO, port = 3000) {
...
+    allowMethod(HttpMethod.Post)
+    allowMethod(HttpMethod.Delete)
+    allowNonSimpleContentTypes = true
+    allowHeader("Mcp-Session-Id")
+    allowHeader("Mcp-Protocol-Version")
+    exposeHeader("Mcp-Session-Id")
+    exposeHeader("Mcp-Protocol-Version")
+}
+```
+
```
devcrocod, line 823: How is it related to PR?

README.md
```diff
@@ -232,10 +228,9 @@ fun main(args: Array<String>) {
         CallToolResult(content = listOf(TextContent("Hello, world!")))
     }
 
+    // mcpStreamableHttp() automatically installs ContentNegotiation with McpJson.
+    // Do NOT install ContentNegotiation yourself — the SDK handles it.
```
devcrocod: Why need this comment in the code snippet?

devcrocod (commented):
Thank you for the PR, this indeed needs to be fixed

However, it is still unclear to me what the user is supposed to do if they want to provide their own JSON configuration
for example, enable pretty printing or set other options in addition to explicitNulls and the rest

Also, in addition to the comments I already left, the AI review pointed out that `KotlinTestBase` was missed as well, and it also uses `install(ContentNegotiation)`

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/LogCapture.kt
kpavlov: DuplicatePluginException should not be thrown, otherwise it will break current users. Also, users should still be able to install their own ContentNegotiation configuration. That's why it is warniong, not an error.
 
The log message that should be verified, this is expected behavior and to verify. When extra dependency and configuration is needed - it is needed

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
kpavlov: Exactly

devcrocod: ?
I mean our extensions: mcp and mcpStreamable

kotlin-sdk-server/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/server/LogCapture.kt
devcrocod: > It's the log message that should be verified. When extra dependency and configuration is needed - it is needed

Why is that? Your comment doesn’t make it clear

README.md
kpavlov: This is a finding from the troubleshooting session. The MCP Inspector does not function with the default configuration without this.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
kpavlov: In this case, we must store an application attribute and make installMcpContentNegotiation() completely idempotent. Done

gradle/libs.versions.toml
kpavlov: For testing, please check the tests.

.github/workflows/build.yml
kpavlov: `build` runs `check` after running tests. Build should fail earlier if it can not be merged.

README.md
devcrocod: Didn't you suggest not mixing unrelated changes in the same PR?

.github/workflows/build.yml
devcrocod: > build runs check after running tests. Build should fail earlier if it can not be merged.

That's not correct. test is part of check

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServerHelpers.kt
kpavlov: Softened

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
kpavlov: done

README.md
kpavlov: I agree, but this often occurs in this project. This note is related to the user’s issue. So "boy scout's rule" apply.

e5l (changes requested):
Two issues found — see inline comments.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServerHelpers.kt
```diff
@@ -0,0 +1,46 @@
...
+    if (pluginOrNull(ContentNegotiation) != null) {
+        // ContentNegotiation was already installed by user code.
+        logger.warn {
+            "ContentNegotiation is already installed. " +
+                "MCP requires json(McpJson) for correct JSON-RPC serialization " +
+                "(explicitNulls = false, encodeDefaults = true). " +
+                "Ensure your ContentNegotiation configuration includes json(McpJson)."
+        }
+    } else {
+        install(ContentNegotiation) {
```
e5l, line 42: When the user has pre-installed `ContentNegotiation` with wrong config (e.g. default `json()` instead of `json(McpJson)`), this logs a warning but doesn't actually fix anything — the runtime serialization failures still happen silently.

This means the most dangerous case (user installed CN *incorrectly*) isn't addressed. Consider either:
- Registering `McpJson` as an additional converter even when CN is already installed
- Failing fast with an exception instead of a warning

As written, the fix only helps users who forgot to install CN entirely, not those who installed it with the wrong config.

kotlin-sdk-server/build.gradle.kts
```diff
@@ -12,7 +12,9 @@ kotlin {
                 api(project(":kotlin-sdk-core"))
                 api(libs.ktor.server.core)
                 api(libs.ktor.server.sse)
+                implementation(libs.ktor.server.content.negotiation)
                 implementation(libs.ktor.server.websockets)
```
e5l, line 16: `ktor-serialization` should already come in transitively via `ktor-server-content-negotiation`. Is there a specific multiplatform reason this is needed as a direct dependency? If not, it can be removed.

kpavlov: Unfortunately, both `ktor-server-content-negotiation` and `ktor-serialization` are required

e5l (approved):
agreed to fix it as separate pr

kpavlov:
Added follow-up issue #668

.github/workflows/build.yml
devcrocod: ktlintCheck doesn't depend on the tests. When running build, the tests and ktlintCheck are executed in parallel. So it is not entirely clear in which scenario "Build should fail earlier" would actually make the build fail faster

Strydom:
Why not improve the documentation instead of installing this automatically. You've now arguably introduced a worse problem https://github.com/modelcontextprotocol/kotlin-sdk/issues/668, where i cannot use the SDK for streaming at all now, i need to fork the logic into my app to I can control the routing and where CN is applied.

PR #670, opened 2026-04-02
feat: server transport state machine, concurrent request handling (#518, #564, #572)

kpavlov:
## Summary

- Introduce `AbstractServerTransport` with `ServerTransportState` state machine — mirrors `AbstractClientTransport` pattern for server-side transports
- Migrate `StdioServerTransport` and `StreamableHttpServerTransport` to use the new state machine (SSE/WebSocket transports untouched — deprecated)
- Launch request handlers concurrently in `Protocol` via supervised coroutine scope — unblocks ping during long tool calls, enables elicitation/sampling during execution
- Structured concurrency: `Protocol.handlerScope` and `StdioServerTransport.scope` inherit the caller's coroutine context (dispatcher, test scheduler) — no hardcoded dispatchers, no detached scopes
- Add `Disconnected`/`Reconnecting` states to `ClientTransportState` (foundation for reconnection in follow-up PR)
- Add `docs/architecture.md` with Mermaid diagrams covering both client and server architecture

‼️👆 See the [architecture document](https://github.com/modelcontextprotocol/kotlin-sdk/blob/kpavlov/transport-lifecycle/docs/architecture.md)

> [!NOTE]
> **The following features are not included and come as follow-up PRs:**
> 1. Protocol reconnection support
> 2. Client reconnection
> 3. Server session recovery on reconnect

## Addresses

- #518 — Unify workflow across server transports
- #564 — Stdio server transport: detached coroutine scope, sequential processing, CancellationException handling
- #572 — Messages should be processed concurrently

## Key changes

**Server transport lifecycle** (`AbstractServerTransport`, `ServerTransportState`):
- State machine: `New → Initializing → Active → ShuttingDown → Stopped`
- `send()` gated on `Active` state, `close()` is idempotent
- Subclasses implement `initialize()`, `performSend()`, `closeResources()`

**Concurrent request execution** (`Protocol`):
- `onRequest()` launches each handler via `handlerScope` (`SupervisorJob` + caller's context)
- `handlerScope` inherits `CoroutineContext` from `connect()` caller — the caller controls the dispatcher; in tests, the test scheduler is inherited automatically
- Handlers launched with `CoroutineStart.UNDISPATCHED` — starts eagerly on caller's thread until first suspension, compatible with `StandardTestDispatcher`
- Active requests tracked in `_activeRequests` for cancellation via `CancelledNotification`
- `StreamableHttpServerTransport` awaits batch completion to keep HTTP connections alive
- Cancellation notification sent on all `CancellationException` paths (not just timeout), delivered via `NonCancellable`

**StdioServerTransport** (#564):
- I/O scope derived from `currentCoroutineContext()` in `initialize()` — structured concurrency, no detached scope
- `CancellationException` always rethrown (verified)

**ClientTransportState**:
- New states: `Disconnected`, `Reconnecting` (wired into transition map and existing tests, methods in follow-up)

**ClientTest fixes**:
- Replaced broken raw `JSONRPCRequest` test with typed `client.listTools()` API
- Fixed timeout test handler that swallowed `CancellationException`
- Added `client.close()`/`server.close()` to all tests to prevent leaked handler scopes

## How Has This Been Tested?

- [x] `ServerTransportStateTest` — valid/invalid transitions, terminal states (24 tests)
- [x] `AbstractServerTransportTest` — lifecycle, send gating, idempotency, error handling (62 tests)
- [x] `StdioServerTransportLifecycleTest` — start/close/send guards (5 tests)
- [x] `StreamableHttpServerTransportLifecycleTest` — start/close/send guards (3 tests)
- [x] `ProtocolConcurrentTest` — concurrent handling, cancellation (2 tests)
- [x] Updated `AbstractClientTransportTest` — new state transitions (92 tests)
- [x] Updated `ServerSessionInitializeTest` — non-deterministic response order
- [x] Updated `ClientTest` — close() calls, typed API, fixed CancellationException handling
- [x] Conformance tests pass (`./conformance-test/run-conformance.sh server`)

### Binary-incompatible (ABI breaks)

1. **`StdioServerTransport` constructor signature unchanged** — was `(Source, Sink)`, remains `(Source, Sink)`. No binary break.

2. **`StdioServerTransport` parent class changed** — `AbstractTransport` → `AbstractServerTransport`. Binary-incompatible for code that references the parent type.

3. **`StreamableHttpServerTransport` parent class changed** — `AbstractTransport` → `AbstractServerTransport`. Same as above.

### Behavioral breaks

4. **`StdioServerTransport.send()` now throws before `start()`** — previously no guard, now throws `McpException(CONNECTION_CLOSED)`. Code that called `send()` without `start()` will now fail.

5. **`StreamableHttpServerTransport.send()` now throws before `start()`** — same as above.

6. **Request handlers now execute concurrently** — previously sequential inline execution in `Protocol.onRequest()`. Code relying on sequential handler execution may see race conditions.

7. **`ClientTransportState.values()` / `entries` changed** — two new enum values (`Disconnected`, `Reconnecting`). Code doing exhaustive `when` without `else` will get a compile error. `ordinal` values shifted for `ShuttingDown`, `ShutdownFailed`, `Stopped`.

### Source-compatible but worth noting

8. **`StdioServerTransport.close()` is no longer re-entrant** — previously toggled `AtomicBoolean(true→false)` allowing close-start-close cycles. Now terminal: `Stopped` is final.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

Addresses #518, #564, #572

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -2,6 +2,28 @@
 
 This document describes the internal architecture of the MCP Kotlin SDK. It covers both client and server sides, the transport abstraction, state management, and concurrency model. 
 
+<!--- TOC -->
+
+* [Layered Design](#layered-design)
+* [Module Structure](#module-structure)
+* [Client Architecture](#client-architecture)
+  * [How the Client Works](#how-the-client-works)
+  * [Client Transport Hierarchy](#client-transport-hierarchy)
+  * [Client Transport State Machine](#client-transport-state-machine)
+* [Server Architecture](#server-architecture)
+  * [How the Server Works](#how-the-server-works)
+  * [Server Class Hierarchy](#server-class-hierarchy)
+  * [Server Transport State Machine](#server-transport-state-machine)
+  * [Server Components](#server-components)
+  * [Session Lifecycle](#session-lifecycle)
+* [Concurrency Model](#concurrency-model)
+  * [Concurrent Request Handling](#concurrent-request-handling)
+  * [Streamable HTTP Request Flow](#streamable-http-request-flow)
+* [Capability Enforcement](#capability-enforcement)
+* [Testing](#testing)
+
+<!--- END -->
+
 ## Layered Design
 
 The SDK follows a four-layer architecture where each layer has a single responsibility and communicates only with its immediate neighbors.
@@ -358,7 +380,68 @@ sequenceDiagram
 
 Without concurrent dispatch, a long-running tool call would block all other request processing, including ping responses. With it, the server (or client) can handle multiple requests in flight simultaneously.
 
-Active request jobs are tracked in `_activeRequests` (keyed by request ID). When a `notifications/cancelled` message arrives, the corresponding job is cancelled. On `close()`, the entire `handlerScope` is cancelled.
+**Context inheritance.** `handlerScope` inherits its `CoroutineContext` from whoever calls `connect()`. The Protocol does not impose a specific dispatcher — the caller controls threading. In production, a server typically calls `connect()` from a Ktor dispatcher or `Dispatchers.IO` scope; in tests, `runTest` provides a `TestCoroutineScheduler` that the handlers inherit automatically. If you need to isolate handler threads from transport I/O, call `connect()` from a dedicated scope (e.g., `Dispatchers.Default.limitedParallelism(16)`), or use `withContext` inside individual handlers.
+
+**Cancellation.** Active request jobs are tracked in `_activeRequests` (keyed by request ID). Two cancellation paths exist:
+
+- **Remote cancellation**: when a `notifications/cancelled` message arrives, the corresponding handler job is cancelled via `_activeRequests[requestId]?.cancel()`.
+- **Local cancellation**: when the calling coroutine is cancelled (e.g., `job.cancel()`), `Protocol.request()` sends a `CancelledNotification` to the remote side (via `NonCancellable` to guarantee delivery even during cancellation), then rethrows.
+
+On `close()`, the entire `handlerScope` is cancelled.
+
+**Example: tool that elicits user input during execution.** Without concurrent dispatch, the elicitation response from the client would be blocked behind the tool handler — a deadlock. With concurrent dispatch, the server handles the elicitation response on a separate coroutine while the tool handler awaits:
+
+```kotlin
+server.addTool("confirm-action", "Asks user to confirm") {
+    // This sends a request TO the client and suspends until the user responds.
+    // Other incoming requests (including the elicitation response) are handled
+    // concurrently by Protocol's handlerScope.
+    val result = createElicitation(
+        message = "Confirm deployment to production?",
+        requestedSchema = ElicitRequestParams.RequestedSchema(
+            properties = mapOf("confirm" to BooleanSchema(description = "Confirm?")),
+        ),
+    )
+    CallToolResult(content = listOf(TextContent(text = "User said: ${result.action}")))
+}
+```
+
+**Example: controlling the handler dispatcher.** The caller of `connect()` decides the threading model:
+
+```kotlin
+// Production: handlers run on a dedicated pool, isolated from transport I/O
+val handlerPool = Dispatchers.Default.limitedParallelism(16)
+CoroutineScope(handlerPool).launch {
+    session.connect(transport)
+}
+
+// Test: handlers inherit the test scheduler — virtual time, no real threads
+@Test
+fun `should handle concurrent requests`() = runTest {
+    client.connect(transport) // handlerScope inherits TestCoroutineScheduler
+    // ...
+}
+```
+
+**Example: cancelling a request.** The client cancels a long-running tool call; the server handler is cancelled cooperatively:
+
+```kotlin
+// Client side
+val job = launch { client.callTool("slow-tool", emptyMap()) }
+// ... later
+job.cancel("User pressed cancel")
+// Protocol.request() sends CancelledNotification to server via NonCancellable,
+// server's _activeRequests[requestId] job is cancelled
+
+// Server side — handler must be cooperative
+server.addTool("slow-tool", "Takes a long time") {
+    repeat(100) { step ->
+        delay(100) // suspension point — CancellationException thrown here
+        reportProgress(step, 100)
+    }
+    CallToolResult(content = listOf(TextContent(text = "Done")))
+}
+```
 
 ```mermaid
 sequenceDiagram
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,6 +1,6 @@
 # Architecture
 
-This document describes the internal architecture of the MCP Kotlin SDK. It covers both client and server sides, the transport abstraction, state management, and concurrency model. All diagrams use Mermaid syntax and render in GitHub, IntelliJ, and most Markdown viewers.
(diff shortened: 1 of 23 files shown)


kpavlov:
@rnett, I’d be grateful if you could take a peek.

docs/architecture.md
```diff
@@ -0,0 +1,527 @@
...
+
+## Concurrency Model
+
+### Concurrent Request Handling
+
+`Protocol.onRequest()` launches each incoming request handler in a supervised coroutine via `handlerScope`. This is essential for MCP features that require bidirectional communication during request handling — for example, a tool call that needs to elicit user input or sample from the LLM while executing.
+
+Without concurrent dispatch, a long-running tool call would block all other request processing, including ping responses. With it, the server (or client) can handle multiple requests in flight simultaneously.
+
+**Context inheritance.** `handlerScope` inherits its `CoroutineContext` from whoever calls `connect()`. The Protocol does not impose a specific dispatcher — the caller controls threading. In production, a server typically calls `connect()` from a Ktor dispatcher or `Dispatchers.IO` scope; in tests, `runTest` provides a `TestCoroutineScheduler` that the handlers inherit automatically. If you need to isolate handler threads from transport I/O, call `connect()` from a dedicated scope (e.g., `Dispatchers.Default.limitedParallelism(16)`), or use `withContext` inside individual handlers.
```
rnett, line 383: While this is certainly nice for test dispatchers, it seems like it would end up with the body of most tool calls being ran in `Dispatchers.IO`, which would not be good.

> > If you need to isolate handler threads from transport I/O, call `connect()` from a dedicated scope (e.g., `Dispatchers.Default.limitedParallelism(16)`), or use `withContext` inside individual handlers.

Neither of these are great solutions: consumers typically don't control where `connect()` is called, e.g. when using Ktor integration; and using `withContext` in every single handler isn't great either.

(this is why I dislike how test dispatchers are dispatchers, but that's a whole nother conversation)

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
```diff
@@ -277,66 +307,87 @@ public abstract class Protocol(@PublishedApi internal val options: ProtocolOptio
...
+        handler: suspend (JSONRPCRequest, RequestHandlerExtra) -> RequestResult?,
+    ) {
+        try {
+            val result = handler(request, RequestHandlerExtra())
+            logger.trace { "Request handled successfully: ${request.method} (id: ${request.id})" }
+            transport?.send(
+                JSONRPCResponse(id = request.id, result = result ?: EmptyResult()),
+            )
+        } catch (_: CancellationException) {
+            // Request cancelled — no response sent
```
rnett, line 369: Shouldn't this be re-thrown?

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
```diff
@@ -277,66 +307,87 @@ public abstract class Protocol(@PublishedApi internal val options: ProtocolOptio
...
-                    ),
-                )
-            } catch (cause: Throwable) {
-                logger.error(cause) { "Error sending method not found response" }
-                onError(cause)
-            }
+            // UNDISPATCHED: start eagerly on the caller's thread until first suspension,
+            // then resume on the scope's dispatcher. This is compatible with StandardTestDispatcher
+            // and avoids requiring a dispatch before the handler starts executing.
+            scope.launch(start = CoroutineStart.UNDISPATCHED) { sendMethodNotFound(request) }
```
rnett, line 328: UNDISPATCHED can still cause requests to block other requests if they do non-suspending blocking before hitting a suspension point, since it would happen in the parent thread (e.g. a `synchronized` block, `Thread.sleep`, etc).  IMO it's acceptable as long as it doesn't cause issues with STDIO, but if we can avoid it we should, and document it if not.

kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
```diff
@@ -277,66 +307,87 @@ public abstract class Protocol(@PublishedApi internal val options: ProtocolOptio
...
+    private suspend fun sendErrorResponse(request: JSONRPCRequest, cause: Throwable) {
+        try {
+            val rpcError = if (cause is McpException) {
+                RPCError(code = cause.code, message = cause.message.orEmpty(), data = cause.data)
+            } else {
+                RPCError(code = RPCError.ErrorCode.INTERNAL_ERROR, message = cause.message ?: "Internal error")
+            }
+            transport?.send(JSONRPCError(id = request.id, error = rpcError))
+        } catch (_: CancellationException) {
+            // Shutting down
```
rnett, line 387: Shouldn't this be re-thrown?

kotlin-sdk-core/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/ProtocolConcurrentTest.kt
```diff
@@ -0,0 +1,135 @@
...
+import io.modelcontextprotocol.kotlin.test.utils.runIntegrationTest
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withTimeout
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import kotlin.time.Duration.Companion.seconds
+
+class ProtocolConcurrentTest {
```
rnett, line 23: This could be a good scenario to use lincheck in, although I'm not that familiar with it.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/AbstractServerTransport.kt
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    @Suppress("TooGenericExceptionCaught", "ThrowsCount")
+    public override suspend fun send(message: JSONRPCMessage, options: TransportSendOptions?) {
+        if (state != ServerTransportState.Active) {
+            throw McpException(
+                code = CONNECTION_CLOSED,
+                message = "Transport is not active",
+            )
+        }
```
rnett, line 107: `ping` and `logging` should be exempt from this: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#:~:text=The%20client%20SHOULD,notification

> The client SHOULD NOT send requests other than [pings](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/ping) before the server has responded to the initialize request.
> The server SHOULD NOT send requests other than [pings](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/ping) and [logging](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging) before receiving the initialized notification.

rnett (commented):
IMO SSE transport is still common enough that it should be updated too, if possible.

kpavlov:
> common enough that it should be updated too, if possible.

Sure, make sense! I just was afraid that it would cause too many changes.

Thank you for swift review, I will address your comments on Monday

PR #706, opened 2026-04-15
fix: propagate CancellationException instead of logging as error

### LINKED ISSUE #242: Cancellation exceptions cause an error to be logged

devcrocod:
fixes #242

## How Has This Been Tested?
new and existing tests pass

## Breaking Changes
none 

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
@@ -24,11 +24,12 @@ import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
-import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 import kotlinx.serialization.SerializationException
 import kotlin.concurrent.atomics.ExperimentalAtomicApi
 import kotlin.time.Duration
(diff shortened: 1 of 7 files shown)


e5l (approved):
lgtm. should we introduce this inspection as part of the code review skill? it is frequent error

devcrocod:
> should we introduce this inspection as part of the code review skill? it is frequent error

This is already documented in agents.md
We could also add a dedicated skill, I'd suggest considering a general-purpose skill for working with coroutines in kotlin-agents-skill

PR #714, opened 2026-04-23
chore: remove idea files related to detekt

devcrocod:
- detekt idea plugin version (1.23.8) is incompatible with the gradle plugin version (2.0.0-alpha.2)
- remove the annoying popup shown when the idea starts 
image


e5l (approved):
lgtm

PR #681, opened 2026-04-09
fix(server): GET SSE stream crash on Netty in Streamable HTTP

devcrocod:
`appendSseHeaders()` was called inside Ktor's `sse {}` block where response headers are already committed. On Netty this throws `UnsupportedOperationException`, killing the SSE stream and causing the client to retry in a loop

## How Has This Been Tested?
unit, simple-streamable-server sample

## Breaking Changes
none

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist
- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x ] I have added appropriate error handling
- [ x] I have added or updated documentation as needed

PR #716, opened 2026-04-24
fix(server): evict stale GET SSE stream mapping on reconnect

### LINKED ISSUE #715: GET SSE stream reconnect fails due to stale STANDALONE_SSE_STREAM_ID mapping
**Describe the bug**

  When a GET SSE stream disconnects and the client reconnects on the same session, the new GET is silently rejected because the previous stream's `STANDALONE_SSE_STREAM_ID` entry in `streamsMapping` hasn't been
  cleaned up yet. Clients see a 200 OK with an immediately-closed empty SSE stream, causing a retry loop that manifests as a permanently lost MCP connection.

  **To Reproduce**

  1. Start a Streamable HTTP MCP server using `mcpStreamableHttp`
  2. Initialize a session via POST to `/mcp`
  3. Open a GET SSE stream on `/mcp`, consume the flush event, then close the connection
  4. Immediately open a new GET SSE stream on `/mcp` with the same session ID
  5. The new stream is dead — first `readLine()` returns null

  **Expected behavior**

  The reconnected GET SSE stream should be live and functional. The TypeScript reference implementation handles this correctly because `ReadableStream.cancel()` removes the stream mapping synchronously on
  disconnect.

  **Logs**

  Server-side `CallLogging` reports 409 Conflict, but this status never reaches the client (Ktor's `sse {}` handler commits 200 OK before the transport handler runs):

```
  13:47:05.141 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=d4aed42f-86af-49da-ab1d-3b8b61f9ad13)
  13:47:05.339 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=9601a9a7-0583-43bb-9c0f-1664c113f771)
  13:47:05.356 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=5092cfe4-c66a-4609-8fde-1d9017307818)
  13:47:06.142 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=d4aed42f-86af-49da-ab1d-3b8b21f9ad13)
  13:47:06.340 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=9601a9a7-0583-43bb-9c0f-1664c113f771)
  13:47:06.358 WARN  io.ktor.server.Application - GET /mcp → 409 Conflict (session=5092cfe4-c66a-4609-8fde-1d9017307818)
```

  Three concurrent MCP client sessions, each retrying ~1/second indefinitely.

  **Additional context**

  The TypeScript reference implementation (`packages/server/src/server/streamableHttp.ts) cleans up the stream mapping synchronously via `ReadableStream.cancel()`:

  ```typescript
  cancel: () => {
      this._streamMapping.delete(this._standaloneSseStreamId);
  }
```

  The Kotlin SDK (StreamableHttpServerTransport.kt line) uses `invokeOnCompletion`, which is asynchronous:

```kotlin
  sseSession.coroutineContext.job.invokeOnCompletion {
      streamsMapping.remove(STANDALONE_SSE_STREAM_ID)
  }
```

  Between the client disconnect and the `invokeOnCompletion` callback firing, there is a race window where `handleGetRequest` finds the stale entry at line 467 and rejects. In practice with Ktor (both CIO and Netty
  engines), this window is long enough that reconnects consistently fail.

  There is a secondary issue: Ktor's `sse {}` handler commits 200 OK headers before calling `handleGetRequest`, so the `call.reject(HttpStatusCode.Conflict, ...)` at line 468-473 can never deliver a 409 to the client.
  The rejection manifests as an empty/closed SSE stream rather than a proper HTTP error response. The 409 in server logs comes from `CallLogging` reading the response object, not the wire.

  This bug likely became reachable after #681 (April 10) fixed the Netty `appendSseHeaders` crash that previously prevented GET SSE from working at all on Netty.

  Tested against latest main at bf8dc6d.

Written with the help of Claude Code Opus 4.7.

amr:
## Motivation and Context

When a GET SSE stream disconnects and the client reconnects on the same session, the new GET is silently rejected because the previous stream's `STANDALONE_SSE_STREAM_ID` entry in `streamsMapping` hasn't been cleaned up yet. Cleanup via `invokeOnCompletion` is asynchronous — there is a race window where the stale entry persists and blocks reconnection.

Additionally, Ktor's `sse {}` handler commits 200 OK headers before `handleGetRequest` runs, so the intended `call.reject(HttpStatusCode.Conflict, ...)` cannot change the already-committed HTTP status. On all tested engines (test host, CIO, Netty), clients receive a 200 OK with an immediately-closed empty SSE stream rather than a 409 Conflict. Server-side `CallLogging` reports 409 (from the response object, not the wire status), which is misleading.

### What this PR does

1. **Explicitly closes the old SSE session** before establishing a new stream. If the old session is already dead, `close()` is a no-op. This is the synchronous cleanup that the async `invokeOnCompletion` couldn't provide.

2. **Replaces `invokeOnCompletion` with `try/finally`** around `awaitCancellation()` — cleanup runs during cancellation propagation (earlier in the lifecycle) rather than after job completion. An identity check (`=== newContext`) ensures only this stream's entry is removed, not a replacement's.

3. **Adds identity-based removal in `emitOnStream`** — a failed `send()` on an old stream can no longer accidentally evict the new stream's mapping.

4. **Rethrows `CancellationException`** instead of swallowing it, preserving structured concurrency.

The TypeScript reference implementation avoids this race via synchronous `ReadableStream.cancel()` cleanup. The Kotlin coroutine equivalent (`invokeOnCompletion`) is inherently async. The Go SDK uses `defer stream.release()` under a mutex for the same reason. The Java SDK uses unconditional `AtomicReference.set()` replacement with `compareAndExchange` on close.

Fixes #715

## How Has This Been Tested?

All tests run against latest `main` (`bf8dc6d`). ktlint and detekt pass clean.

- 2 new unit tests (Ktor test host): reconnect after disconnect, concurrent stream takeover
- 2 new integration tests (real CIO embedded server): same scenarios over real HTTP
- Manual verification against a live Ktor+Netty MCP server (curl): confirmed 200 OK on the wire with immediately-closed stream on reconnect (pre-fix), live stream on reconnect (post-fix)
- All existing server unit tests pass (76 total)
- All existing integration tests pass (306 total)

## Breaking Changes

The server no longer returns 409 Conflict for a second GET SSE stream on the same session. Instead, the old stream is explicitly closed and the new stream takes over. The 409 was already unreachable as an HTTP status — Ktor's `sse {}` handler commits 200 OK before the transport handler runs on all tested engines (test host, CIO, Netty), so clients never received 409 on the wire.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

This bug likely became reachable after #681 (April 10) fixed the Netty `appendSseHeaders` crash that previously prevented GET SSE from working at all on Netty.

Cross-SDK comparison of how each SDK handles a second GET SSE on the same session:

| SDK | Behavior | Cleanup mechanism | Stale mapping bug? |
|-----|----------|-------------------|-------------------|
| TypeScript (reference) | 409 Conflict | Synchronous `ReadableStream.cancel()` | No |
| Go | 409 Conflict | `defer stream.release()` under mutex | No |
| Java | Replaces (no 409) | `AtomicReference.set()` + `compareAndExchange` on close | No |
| C# | 400 Bad Request | Boolean flag, never reset (no reconnect support) | No |
| Kotlin (before fix) | 409 (unreachable on wire) | Async `invokeOnCompletion` | **Yes** |
| Kotlin (this PR) | Closes old + replaces | `try/finally` + identity-based removal | No |

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
@@ -464,27 +464,43 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
             }
         }
 
-        if (STANDALONE_SSE_STREAM_ID in streamsMapping) {
-            call.reject(
-                HttpStatusCode.Conflict,
-                RPCError.ErrorCode.CONNECTION_CLOSED,
-                "Conflict: Only one SSE stream is allowed per session",
-            )
-            return
+        streamsMapping[STANDALONE_SSE_STREAM_ID]?.let { existingContext ->
+            // Close the previous SSE session. If the stream is already dead this
+            // is a no-op. If it is still alive, closing it cancels the coroutine
+            // blocked in awaitCancellation(), which triggers the identity-guarded
+            // finally block to remove the mapping.
+            try {
+                existingContext.session?.close()
+            } catch (_: CancellationException) {
+                throw CancellationException("Cancelled while closing previous SSE stream")
+            } catch (_: Exception) {
+                // Ignore — the old stream may already be closed.
+            }
+            // After closing, give the old coroutine's finally block a chance to
+            // remove the mapping. If the entry is still present (race edge case),
+            // evict it — the old session is closed either way.
+            streamsMapping.remove(STANDALONE_SSE_STREAM_ID)
         }
 
         // SSE headers (Content-Type, Cache-Control, Connection) are already set by the framework's SSE handler
         flushSse(sseSession)
-        streamsMapping[STANDALONE_SSE_STREAM_ID] = SessionContext(sseSession, call)
+        val newContext = SessionContext(sseSession, call)
+        streamsMapping[STANDALONE_SSE_STREAM_ID] = newContext
         val clientProtocolVersion =
             call.request.header(MCP_PROTOCOL_VERSION_HEADER) ?: DEFAULT_NEGOTIATED_PROTOCOL_VERSION
         maybeSendPrimingEvent(STANDALONE_SSE_STREAM_ID, sseSession, clientProtocolVersion)
-        sseSession.coroutineContext.job.invokeOnCompletion {
-            streamsMapping.remove(STANDALONE_SSE_STREAM_ID)
-        }
         // Keep the SSE connection open until the client disconnects or the transport is closed.
-        // Without this, the Ktor sse{} handler returns immediately, closing the stream.
-        awaitCancellation()
+        // Cleanup uses try/finally (runs during cancellation propagation) instead of
+        // invokeOnCompletion (runs after job completion) to minimize the window between
+        // disconnect and mapping removal. Identity check ensures only this stream's entry
+        // is removed — not a replacement that arrived in the meantime.
+        try {
+            awaitCancellation()
+        } finally {
+            if (streamsMapping[STANDALONE_SSE_STREAM_ID] === newContext) {
+                streamsMapping.remove(STANDALONE_SSE_STREAM_ID)
+            }
+        }
     }
 
     /** Handles an HTTP DELETE request by closing the session and the transport. */
@@ -725,10 +741,17 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
         try {
             session?.send(event = "message", id = eventId, data = McpJson.encodeToString(message))
         } catch (e: CancellationException) {
-            streamsMapping.remove(streamId)
+            // Identity-based removal: only evict this stream's entry, not a replacement's.
+            val current = streamsMapping[streamId]
+            if (current?.session === session) {
+                streamsMapping.remove(streamId)
+            }
             throw e
         } catch (_: Exception) {
-            streamsMapping.remove(streamId)
+            val current = streamsMapping[streamId]
+            if (current?.session === session) {
+                streamsMapping.remove(streamId)
+            }
         }
     }
 
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
@@ -464,28 +464,26 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
             }
         }
 
-        streamsMapping[STANDALONE_SSE_STREAM_ID]?.let { existingContext ->
-            // Close the previous SSE session. If the stream is already dead this
-            // is a no-op. If it is still alive, closing it cancels the coroutine
-            // blocked in awaitCancellation(), which triggers the identity-guarded
-            // finally block to remove the mapping.
-            try {
-                existingContext.session?.close()
-            } catch (_: CancellationException) {
-                throw CancellationException("Cancelled while closing previous SSE stream")
-            } catch (_: Exception) {
-                // Ignore — the old stream may already be closed.
-            }
-            // After closing, give the old coroutine's finally block a chance to
-            // remove the mapping. If the entry is still present (race edge case),
-            // evict it — the old session is closed either way.
-            streamsMapping.remove(STANDALONE_SSE_STREAM_ID)
-        }
-
         // SSE headers (Content-Type, Cache-Control, Connection) are already set by the framework's SSE handler
         flushSse(sseSession)
         val newContext = SessionContext(sseSession, call)
-        streamsMapping[STANDALONE_SSE_STREAM_ID] = newContext
+        streamMutex.withLock {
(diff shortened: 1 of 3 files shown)


kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -464,13 +464,15 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
             }
         }
 
-        if (STANDALONE_SSE_STREAM_ID in streamsMapping) {
-            call.reject(
-                HttpStatusCode.Conflict,
-                RPCError.ErrorCode.CONNECTION_CLOSED,
-                "Conflict: Only one SSE stream is allowed per session",
-            )
-            return
```
devcrocod, line 473: By removing the reject with a 409 status here, you are changing the contract

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -464,13 +464,15 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
...
+        streamsMapping.remove(STANDALONE_SSE_STREAM_ID)?.let { existingStream ->
+            // Close the previous SSE stream. This handles both stale streams
+            // (client disconnected but cleanup hasn't fired) and live streams
+            // (client opened a new connection, replacing the old one).
+            try {
+                existingStream.session?.close()
+            } catch (_: Exception) {
```
devcrocod: rethrow CancellationException

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -464,13 +464,15 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
...
         }
 
-        if (STANDALONE_SSE_STREAM_ID in streamsMapping) {
-            call.reject(
-                HttpStatusCode.Conflict,
-                RPCError.ErrorCode.CONNECTION_CLOSED,
-                "Conflict: Only one SSE stream is allowed per session",
-            )
-            return
+        streamsMapping.remove(STANDALONE_SSE_STREAM_ID)?.let { existingStream ->
```
devcrocod: Since there is no longer a return, execution reaches `invokeOnCompletion`, and as I understand it, this may introduce a race condition

devcrocod (changes requested):
Thank you for investigating the bug and for your contribution

In the PR, you point out a difference from ts sdk implementation, but this PR changes the behavior contract in case of a conflict. 
I think it would be better to make cleanup synchronous instead. This way, we can preserve the existing contract while fixing the race condition between the old invokeOnCompletion and the new put. WDYT?

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
amr: Thanks for taking a look @devcrocod 

I agree with you, that was my first intuition but after diving in I learned:

1. The actual / "on the wire" behavior of the Kotlin SDK today is that it doesn't actually return 409, the client  gets a 200 (see additional context section + tests)
2. I couldn't actually find a reference to this 409 requirement in the spec, but I did find it in the typescript implementation, so I ran a comparison on 3 more SDKs: Java, Go & C#, and I found that:
    * Java: Replaces stream (no 409)
    * Go: 409 Conflict
    * C#: 400 Bad Request

devcrocod: true

It isn't explicitly stated, but we try to follow the TypeScript and Python SDKs specifically. They are the original SDKs, and most servers/clients are built around them, so we pay close attention to their behavior

devcrocod (approved):
lgtm

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/StreamableHttpServerTransport.kt
```diff
@@ -725,10 +738,17 @@ public class StreamableHttpServerTransport(private val configuration: Configurat
         try {
             session?.send(event = "message", id = eventId, data = McpJson.encodeToString(message))
         } catch (e: CancellationException) {
-            streamsMapping.remove(streamId)
+            // Atomic compare-and-remove — only evict this stream's entry, not a replacement's.
+            val current = streamsMapping[streamId]
```
e5l, line 742: would be nice to extract common logic for catch blocks

e5l (approved):
lgtm

PR #733, opened 2026-04-29
feat(server): expose client implementation and capabilities to handlers

### LINKED ISSUE #552: Make client information available in server handlers

It is not possible for server handlers (e.g. tools) to react to information about the client that is connected.  Information that is useful includes:
 * The client's name & version (aka `Implementation`) - useful for patching around specific client behaviors, if some are known to behave poorly or need special handling.
 * The client capabilities.  This is required to know which messages can be sent without resulting in unexpected exceptions or erroneous handling. Graceful fallback is often required when certain capabilities are not supported by clients. It's not really possible to do this currently - calling non-supported methods results in non-specific `IllegalStateException`s, which could have come from anywhere.

While server capabilities could also be helpful, it's something that you control anyways, so it's much less impactful.

Add `clientImplementation` and `clientCapabilities` to `ClientConnection`.

There's a workaround, which is to use the session ID to extract the full `ServerSession` object from the `Server` instance.  But that's ugly and awkward. 

**Additional context**

jskjw157:
Adds `clientImplementation` and `clientCapabilities` properties on `ClientConnection`, delegating to what `ServerSession` already tracks during the initialize handshake.

closes #552

## Motivation and Context
Server handlers (tools, prompts, resources) currently have no first-class way to read the connected client's name/version or advertised capabilities. The only workaround is to extract the full `ServerSession` from the `Server` instance via `sessionId` — awkward, and easy to get wrong. This blocks two real use cases:

- Adapting handler behavior for known client-side quirks
- Gracefully falling back when a capability (sampling, roots, elicitation) is not advertised, instead of letting `createMessage`/`listRoots`/`createElicitation` fail with a non-specific `IllegalStateException`

## How Has This Been Tested?
Integration tests in `ClientConnectionTest` assert both properties are populated inside tool, prompt, and resource handlers (matching the existing `callAllMethods` pattern that exercises every `ClientConnection` member).

```
./gradlew :integration-test:jvmTest --tests "*ClientConnectionTest"
./gradlew :kotlin-sdk-server:apiCheck
./gradlew :kotlin-sdk-server:detekt :integration-test:detekt
./gradlew :integration-test:ktlintCheck
```

## Breaking Changes
None for consumers. Adds two abstract members to `ClientConnection`, which would require any external implementer to recompile — but `ClientConnectionImpl` is the only implementation and is `internal`.

A small naming note: `clientImplementation` matches the MCP spec's `clientInfo: Implementation` field as suggested in the issue. This creates temporary dual naming with the existing `ServerSession.clientVersion` (also an `Implementation`). Renaming `ServerSession.clientVersion` is a wider public-API change, so I left it as a follow-up unless maintainers want it bundled here.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
@@ -2,6 +2,7 @@ package io.modelcontextprotocol.kotlin.sdk.server
 
 import io.github.oshai.kotlinlogging.KotlinLogging
 import io.modelcontextprotocol.kotlin.sdk.shared.RequestOptions
+import io.modelcontextprotocol.kotlin.sdk.types.ClientCapabilities
 import io.modelcontextprotocol.kotlin.sdk.types.CreateMessageRequest
 import io.modelcontextprotocol.kotlin.sdk.types.CreateMessageResult
 import io.modelcontextprotocol.kotlin.sdk.types.ElicitRequest
@@ -11,6 +12,7 @@ import io.modelcontextprotocol.kotlin.sdk.types.ElicitRequestURLParams
 import io.modelcontextprotocol.kotlin.sdk.types.ElicitResult
 import io.modelcontextprotocol.kotlin.sdk.types.ElicitationCompleteNotification
 import io.modelcontextprotocol.kotlin.sdk.types.EmptyResult
+import io.modelcontextprotocol.kotlin.sdk.types.Implementation
 import io.modelcontextprotocol.kotlin.sdk.types.IncludeContext
 import io.modelcontextprotocol.kotlin.sdk.types.ListRootsRequest
 import io.modelcontextprotocol.kotlin.sdk.types.ListRootsResult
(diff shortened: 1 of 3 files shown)


jskjw157:
@devcrocod when you get a chance — implements #552 by delegating to what `ServerSession` already exposes. The one open question is whether to consolidate the `clientImplementation` / `ServerSession.clientVersion` naming in this PR or leave it as a follow-up; happy either way.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -41,6 +43,19 @@ public interface ClientConnection {
      */
     public val sessionId: String
 
+    /**
+     * The client's reported [Implementation] (name and version) after initialization.
+     */
+    public val clientImplementation: Implementation?
```
devcrocod, line 49: These could be non-null properties

devcrocod (commented):
As with pr #737, the purpose of `ClientConnection` is not entirely clear, since properties are being added here as well
This significantly complicates the class itself and makes its responsibility less clear

devcrocod:
Thanks @jskjw157!
Closing this for the same reason as #737, we don't want to grow `ClientConnection` with client-metadata responsibility
the need stays tracked in #552 and we'll revisit the surface.

PR #734, opened 2026-04-30
fix: require status field to discriminate GetTaskResult from GetTaskPayloadResult

### LINKED ISSUE #601: type-unsafe cast in Protocol result deserialization

blackwell-systems:
## Motivation and Context

Fixes #601

The content-based polymorphic deserializer in `selectClientResultDeserializer` and `selectServerResultDeserializer` matches `GetTaskResult` on the presence of `"taskId"` alone. This incorrectly captures `GetTaskPayloadResult` responses (`tasks/result`), which wrap arbitrary JSON and may also contain a `"taskId"` key but not a `"status"` key.

When `Protocol.request<GetTaskPayloadResult>()` receives a response that was deserialized as `GetTaskResult`, the unchecked `as T` cast at line 478 of Protocol.kt throws a `ClassCastException`.

## Fix

Tighten the discriminator in both `selectClientResultDeserializer` and `selectServerResultDeserializer` to require both `"taskId"` and `"status"`, since `GetTaskResult` always has both fields (status is required per the MCP task schema). `GetTaskPayloadResult` responses (which lack `"status"`) now correctly fall through to other deserializers or the empty result handler.

## How Has This Been Tested?

Added two test cases to `TasksTest`:

1. **`task payload result with content field is not deserialized as GetTaskResult`**: verifies that a `CallToolResult`-shaped JSON (with `"content"` but no `"status"`) is correctly deserialized as `CallToolResult`, not `GetTaskResult`
2. **`task get result with status is correctly deserialized as GetTaskResult`**: verifies that JSON with both `"taskId"` and `"status"` is still correctly deserialized as `GetTaskResult`

All 37 existing tests continue to pass.

## Types of changes

## Checklist

### DIFF
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/types/serializers.kt
@@ -415,7 +415,11 @@ private fun selectClientResultDeserializer(element: JsonElement): Deserializatio
         "action" in jsonObject -> ElicitResult.serializer()
         "task" in jsonObject -> CreateTaskResult.serializer()
         "tasks" in jsonObject -> ListTasksResult.serializer()
-        "taskId" in jsonObject -> GetTaskResult.serializer()
+        // GetTaskResult always has both taskId and status. Matching on taskId alone
+        // would incorrectly capture GetTaskPayloadResult responses (tasks/result),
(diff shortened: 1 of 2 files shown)


devcrocod:
This PR doesn't fix #601 
the test added in PR doesn't show regression, it passes even without changes 
to fix the issue, the request handling itself needs to be corrected, and unchecked cast should be removed 

You are welcome to reopen the PR once it includes a fix that directly addresses the issue and adds a test that fails without the fix and passes with it.

PR #737, opened 2026-04-30
feat(server): add onSessionClose hook to ClientConnection

### LINKED ISSUE #553: Add a way to register session cleanup actions from server handlers

Some tools need to create long-lived resources, that may be queried in the future by other tools or interactions. Currently, it is impossible to clean these resources up when a MCP session ends.

Add a `onSessionClose` method to `ClientConnection`, that adds the close handler to the `ServerSession`.

This can be somewhat worked around by using the session ID to get the session from the server, but it's ugly and awkward.  Even the server isn't easily accessible from the handlers.

**Additional context**

jskjw157:
Adds an `onSessionClose(block: () -> Unit)` hook on `ClientConnection`, delegating to the existing callback chain on `ServerSession.onClose`.

closes #553

## Motivation and Context
Tool, prompt, and resource handlers often acquire session-scoped resources — DB connections, temporary files, background coroutines — that need to be released when the session terminates. Today there's no first-class way for a handler to register that cleanup: the only workaround is to fish the `ServerSession` out of the `Server` via `sessionId` and call its internal `onClose`, which is awkward and easy to get wrong.

This mirrors the access pattern from #552 (#733): expose what `ServerSession` already does through the public `ClientConnection` surface so handlers can reach it directly.

## How Has This Been Tested?
Two integration tests in `ClientConnectionTest`:

- `onSessionClose callback runs when the session closes` — register from a tool handler, close the client, assert the callback fires.
- `multiple onSessionClose callbacks run in registration order` — verifies the documented ordering guarantee inherited from `ServerSession.onClose`.

Only the tool-handler case is exercised because `onSessionClose` is session-scoped — the registration site (tool vs. prompt vs. resource) doesn't change behavior, and adding three near-identical tests would just be noise.

```
./gradlew :integration-test:jvmTest --tests "*ClientConnectionTest"
./gradlew :kotlin-sdk-server:apiCheck
./gradlew :kotlin-sdk-server:detekt
./gradlew ktlintCheck
```

## Breaking Changes
None for consumers. Adds one abstract method to `ClientConnection`, which would require any external implementer to recompile — but `ClientConnectionImpl` is the only implementation and is `internal`.

## Types of changes
- [ ] Documentation update

## Checklist


jskjw157:
@devcrocod follow-up to #733 — same access pattern, just delegating to the existing `ServerSession.onClose` chain. Happy to adjust the API shape (e.g. naming, returning a deregistration handle) if you'd prefer something different.

kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/ClientConnection.kt
```diff
@@ -303,6 +312,10 @@ internal class ClientConnectionImpl(private val session: ServerSession) : Client
         notification(notification)
     }
 
+    override fun onSessionClose(block: () -> Unit) {
```
devcrocod, line 315: If a callback is registered after the session has already been closed, it will never be invoked

devcrocod, line 315: It seems like this could potentially cause a leak. Would it make sense to use DisposableHandle?

devcrocod (commented):
I have some doubts
On the one hand, this is a useful method. On the other hand, it significantly blurs the responsibility of `ClientConnection`, since it adds lifecycle management to it

devcrocod:
Thanks @jskjw157!
Closing this, we'd rather not extend `ClientConnection` with session-lifecycle responsibility
the need stays tracked in #553 and we'll revisit it with a dedicated session-scoped API

PR #738, opened 2026-05-01
fix(client): invoke onClose when SSE session disconnects (#437)

### LINKED ISSUE #437: SseClientTransport isn't closed when underlying sseSession is disconnected

jskjw157:
Fixes the SSE client transport so that registered `onClose` callbacks fire when the underlying SSE session ends — whether the server closes the stream gracefully or an error event terminates it.

closes #437

## Motivation and Context
`SseClientTransport.collectMessages` already runs `closeResources()` in its `finally` block when `session.incoming` completes, but it never invoked the `onClose` callback chain. As a result, the MCP `Client` (and any user code calling `transport.onClose { ... }`) had no way to detect that the server-side SSE connection had dropped.

The other client transports already do this correctly:
- `StdioClientTransport` calls `invokeOnCloseCallback()` from its read-loop `finally` (kotlin-sdk-client/src/.../StdioClientTransport.kt:224).
- The protected `AbstractTransport.invokeOnCloseCallback()` helper guards against duplicate firing with an `AtomicBoolean`, so it is safe to call from both the read-loop `finally` and the user-initiated `close()` path.

The fix mirrors that established pattern in one line.

## How Has This Been Tested?
Added `onClose callback fires when the SSE stream is disconnected by the server` to `SseClientTransportTest`. The test:

1. Starts the transport against a mock SSE engine.
2. Closes the SSE byte channel server-side via a new `MockSseClientEngine.disconnectSseStream()` helper.
3. Asserts the registered `onClose` callback fires (using Kotest `eventually`, since the close propagates on a real-time dispatcher).

I also verified the test fails before the fix and passes after.

```
./gradlew :kotlin-sdk-client:jvmTest
./gradlew apiCheck ktlintCheck detekt
```

## Breaking Changes
None. `invokeOnCloseCallback()` is `protected` on `AbstractTransport`; this PR only adds a call site inside `SseClientTransport`. No public API surface changes.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
@@ -136,6 +136,7 @@ public class SseClientTransport(
             throw e
         } finally {
             closeResources()
+            invokeOnCloseCallback()
         }
     }
 
--- a/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/MockSseClientEngine.kt
+++ b/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/MockSseClientEngine.kt
@@ -53,6 +53,14 @@ internal open class MockSseClientEngine(
         job.cancel()
     }
 
+    /**
+     * Closes the SSE byte channel without disposing the engine, simulating a server-side
+     * disconnect of the SSE stream.
+     */
+    fun disconnectSseStream() {
+        channel.close()
+    }
+
     private fun responseContext(): CoroutineContext = dispatcher + Job()
 
     private suspend fun respondWithSseEndpointEvent(data: HttpRequestData): HttpResponseData {
--- a/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/SseClientTransportTest.kt
+++ b/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/SseClientTransportTest.kt
@@ -1,5 +1,6 @@
 package io.modelcontextprotocol.kotlin.sdk.client.sse
 
+import io.kotest.assertions.nondeterministic.eventually
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
 import io.ktor.client.HttpClient
@@ -9,6 +10,7 @@ import io.modelcontextprotocol.kotlin.sdk.client.SseClientTransport
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCNotification
 import kotlinx.coroutines.test.runTest
 import kotlin.test.Test
+import kotlin.time.Duration.Companion.seconds
 
 class SseClientTransportTest {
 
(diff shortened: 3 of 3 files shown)


jskjw157:
@devcrocod when you have a moment — small fix for the SSE disconnect-detection bug from #437. The change is one line plus a regression test; the existing StdioClientTransport already does this, so this aligns SSE with the established pattern. Happy to adjust the test setup if you prefer a different approach to simulating the server-side disconnect.

devcrocod (approved):
thanks for your contribution
lgtm

Issue #344, opened 2025-10-24
SSE needs a heartbeat


rustyrazorblade:
Some clients (claude code for example) close SSE servers unless there's a heartbeat.  I found the Kotlin MCP server to be extremely unreliable unless it was set up.

```
mcp {
  heartbeat {
    period = 1.seconds
    event = ServerSentEvent("heartbeat")
  }
  server
}
```

UnscientificJsZhai:
Here is how I handled it. While this might not be a direct replacement for a proper SSE heartbeat, it successfully prevented the client from disconnecting in my specific use case. (Gemini CLI)

```kotlin
    embeddedServer(/* ... */) {
        /* ... */
        launch {
            while (true) {
                delay(30.seconds)
                mcpServer?.sessions?.forEach { (_, session) ->
                    try {
                        session.transport?.send(
                            JSONRPCNotification(method = Method.Custom("heartbeat").value, params = null)
                        )
                    } catch (e: Exception) {
                        logger.warn("Sending heartbeat error", e)
                    }
                }
            }
        }
    }
```

PR #761, opened 2026-05-11
feat: SSE heartbeat

UnscientificJsZhai:
## Summary

Adds optional SSE heartbeat configuration for streamable HTTP MCP server connections.

This lets `Application.mcpStreamableHttp` callers pass Ktor's `Heartbeat` configuration for SSE streams, including heartbeat period and event payload. Heartbeats remain disabled by default, preserving the existing stream behavior unless the new option is explicitly provided.

## Motivation and Context

Some MCP clients need heartbeat messages on SSE connections to keep the connection alive. Without those heartbeats, the client may proactively disconnect from the MCP server. MCP server developers need a way to configure an SSE heartbeat mechanism so they can avoid unexpected client disconnects.

In my case, my MCP server was consistently disconnected by Gemini CLI. I worked around the issue with some fallback approaches, but I believe the best implementation is to add heartbeat support by using Ktor's built-in API.

## Implementation Notes

- Added a function overload of `Application.mcpStreamableHttp` to introduce heartbeat parameter `sseHeartbeatConfig: (Heartbeat.() -> Unit)?`.
- Applied the configuration inside the Ktor `sse { ... }` block before handling the existing streamable HTTP transport request.
- Updated the server API dump for the new public API surface.

## How Has This Been Tested?

Added `StreamableHttpHeartbeatIntegrationTest` covering:

- Configured GET SSE streams apply the provided heartbeat configuration.
- GET SSE streams do not send default heartbeats when `sseHeartbeatConfig` is omitted.
- Configured GET SSE streams emit heartbeat events repeatedly.
- Legacy MCP over SSE APIs do not expose the new heartbeat configuration.

Verified locally with:

```bash
./gradlew :integration-test:jvmTest --tests "io.modelcontextprotocol.kotlin.sdk.integration.streamablehttp.StreamableHttpHeartbeatIntegrationTest"
```

Result: `BUILD SUCCESSFUL`.

## Breaking Changes

None.

This change adds a new `Application.mcpStreamableHttp` overload that accepts
`sseHeartbeatConfig: (Heartbeat.() -> Unit)?` while keeping the existing overload
unchanged. Existing callers continue to compile and keep the previous behavior:
SSE heartbeats are not sent unless the new overload is used with an explicit
heartbeat configuration.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

Heartbeat behavior is delegated to Ktor's SSE heartbeat support, so applications can use the same configuration semantics they would use in native Ktor SSE routes.

Related issue: [SSE needs a heartbeat #344 ](https://github.com/modelcontextprotocol/kotlin-sdk/issues/344)

### DIFF
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/streamablehttp/AbstractStreamableHttpIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/streamablehttp/AbstractStreamableHttpIntegrationTest.kt
@@ -3,6 +3,7 @@ package io.modelcontextprotocol.kotlin.sdk.integration.streamablehttp
 import io.ktor.server.cio.CIOApplicationEngine
 import io.ktor.server.engine.EmbeddedServer
 import io.ktor.server.engine.embeddedServer
+import io.ktor.server.sse.Heartbeat
 import io.modelcontextprotocol.kotlin.sdk.server.Server
 import io.modelcontextprotocol.kotlin.sdk.server.ServerOptions
 import io.modelcontextprotocol.kotlin.sdk.server.mcpStreamableHttp
@@ -17,7 +18,10 @@ import io.ktor.server.cio.CIO as ServerCIO
 
 open class AbstractStreamableHttpIntegrationTest {
 
-    suspend fun initTestServer(name: String? = null): StreamableHttpTestServer {
+    suspend fun initTestServer(
+        name: String? = null,
+        sseHeartbeatConfig: (Heartbeat.() -> Unit)? = null,
+    ): StreamableHttpTestServer {
         val mcpServer = Server(
             Implementation(name = name ?: DEFAULT_SERVER_NAME, version = VERSION),
             ServerOptions(
@@ -52,7 +56,7 @@ open class AbstractStreamableHttpIntegrationTest {
             host = URL,
             port = PORT,
         ) {
-            mcpStreamableHttp { mcpServer }
+            mcpStreamableHttp(sseHeartbeatConfig = sseHeartbeatConfig) { mcpServer }
         }
 
         return StreamableHttpTestServer(
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/streamablehttp/StreamableHttpHeartbeatIntegrationTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/integration/streamablehttp/StreamableHttpHeartbeatIntegrationTest.kt
@@ -1,0 +1,174 @@
+package io.modelcontextprotocol.kotlin.sdk.integration.streamablehttp
+
+import io.kotest.matchers.shouldBe
+import io.ktor.client.HttpClient
+import io.ktor.client.plugins.sse.SSE
+import io.ktor.client.request.header
+import io.ktor.client.request.post
+import io.ktor.client.request.prepareGet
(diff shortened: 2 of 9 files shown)


devcrocod (changes requested):
This PR adds the ktor api to the transport layer, which we would like to avoid since we plan to move away from a tight coupling with ktor.
Please consider using heartbeat only in the ktor dsl

UnscientificJsZhai:
> This PR adds the ktor api to the transport layer, which we would like to avoid since we plan to move away from a tight coupling with ktor. Please consider using heartbeat only in the ktor dsl

The problem is that we can't access `io.ktor.server.sse.sse` block from outside `io.modelcontextprotocol.kotlin.sdk.server.mcpStreamableHttp` block. 

`KtorServer.kt` is a file contains a lot of Ktor API endpoints, It is already tightly coupled with Ktor. 

Maybe I should remove the reference of `io.ktor.server.sse.Heartbeat` from `StreamableHttpServerTransport.kt‎` and use the parameters we defined? However, this would limit the user's ability to control the SSE heartbeat.

devcrocod:
> The problem is that we can't access io.ktor.server.sse.sse block from outside io.modelcontextprotocol.kotlin.sdk.server.mcpStreamableHttp block.

Maybe I'm missing something, but couldn't we just pass an additional parameter and use the heartbeat in serverSSESession? That way, we wouldn't have to expose the ktor api in core.

UnscientificJsZhai:
> > The problem is that we can't access io.ktor.server.sse.sse block from outside io.modelcontextprotocol.kotlin.sdk.server.mcpStreamableHttp block.
> 
> Maybe I'm missing something, but couldn't we just pass an additional parameter and use the heartbeat in serverSSESession? That way, we wouldn't have to expose the ktor api in core.

I overthought this. I initially thought I needed to extract the mandatory SSE heartbeat parameters (like the interval) into a new configuration class to avoid a direct dependency on Ktor's config.

However, since `KtorServer.kt` is already tightly coupled with Ktor anyway, it makes much more sense to handle everything right there.

UnscientificJsZhai:
To ensure binary compatibility for the API, I opted to add a function overload to introduce this parameter.
```kotlin
@KtorDsl
public fun Application.mcpStreamableHttp(
    path: String = "/mcp",
    enableDnsRebindingProtection: Boolean = true,
    allowedHosts: List<String>? = null,
    allowedOrigins: List<String>? = null,
    eventStore: EventStore? = null,
    sseHeartbeatConfig: (Heartbeat.() -> Unit)?,
    block: RoutingContext.() -> Server,
) { /* ... */ }
```
I plan to mark the old function as deprecated later and set a default value `null` for the new parameter in the new function.

Please let me know if a breaking change like this to the API would actually be preferable in this scenario:
```kotlin
@KtorDsl
public fun Application.mcpStreamableHttp(
    path: String = "/mcp",
    enableDnsRebindingProtection: Boolean = true,
    allowedHosts: List<String>? = null,
    allowedOrigins: List<String>? = null,
    eventStore: EventStore? = null,
    sseHeartbeatConfig: (Heartbeat.() -> Unit)? = null,
    block: RoutingContext.() -> Server,
) { /* ... */ }
```

devcrocod (changes requested):
Sorry for the long wait
Overall I like this solution, but regarding binary compatibility, please remove the overload. Unfortunately, this won't work because Kotlin won't be able to resolve the function. So add a new default parameter to the old method. This is a source-compatible solution, so it is fine for now while the library is unstable.
Also, please remove the `legacy MCP over SSE APIs do not expose heartbeat configuration` test, this test is redundant

after that I will gladly merge your changes

devcrocod (approved):
lgtm
thanks so much for the contribution!

PR #798, opened 2026-06-04
fix(client): close stdio read sources during shutdown

### LINKED ISSUE #514: StdioClientTransport fails to close gracefully when Java stream hangs in blocking read
# Description

**Affected version: 0.8.3**

`StdioClientTransport.readSource()` uses `source.readAtMostTo(buffer, BUFFER_SIZE)`, which blocks the thread when reading from Java streams (e.g., `ProcessBuilder().start().inputStream.asSource()`).

When close() is called:
  1. Scope gets cancelled via `scope.cancel()`
  2. Read coroutine remains blocked in `readAtMostTo()`
  3. Cancellation check `while (isActive)` only runs after the read completes
  4. If no data arrives and the stream isn't closed, transport never exits cleanly

### Impact
  - Transport cannot shut down until data arrives or the stream closes
  - Thread leakage in hanging scenarios
  - Tests requiring a clean shutdown must explicitly close all pipes

### Root Cause
  kotlinx.io's blocking I/O API doesn't respect coroutine cancellation. The underlying Java InputStream.read() cannot be interrupted from the
  coroutine layer.

### To Reproduce
Can be reproduced [in this test](https://github.com/modelcontextprotocol/kotlin-sdk/blob/7eb2a93c466bf91d0a4932d736ddfe8d1e60b8f9/integration-test/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/BaseTransportTest.kt#L30)

## Expected behaviour

`Transport.close()` cancels the coroutine and releases InputStream

## Additional context

Workaround Attempted

```kotlin
private suspend fun CoroutineScope.readSource(...) {                                                                                    
   while (isActive) {                                                                                                                   
     val bytesRead = try {                                                                                                              
       withTimeout(READ_TIMEOUT_MS) {                                                                                                   
         source.readAtMostTo(buffer, BUFFER_SIZE) // blocking here!                                                                      
       }                                                                                                                                
     } catch (e: TimeoutCancellationException) {                                                                                        
       yield() // Check cancellation, then retry                                                                                        
       continue                                                                                                                         
     }                                                                                                                                  
     // ... process bytes                                                                                                               
   }                                                                                                                                    
 } 
```

Using withTimeout + yield() provides periodic cancellation checks but does not interrupt the blocked read operation—it only detects cancellation between timeout cycles.

**Note:** This is the best achievable with the current kotlinx.io blocking APIs. Any fix will have inherent limitations unless switching to platform-specific non-blocking I/O.

### Constraints

This is a fundamental limitation of blocking I/O with coroutines. True solution requires:
  - Non-blocking I/O (not available in kotlinx.io for JVM)
  - Platform-specific interruption mechanisms (e.g., Thread.interrupt())
  - Redesigning transport architecture to use non-blocking primitives

jstar0:
## Summary

Fixes `StdioClientTransport.close()` so shutdown does not hang when the stdin read loop is blocked waiting for input.

## Root Cause / Context

`StdioClientTransport.closeResources()` stopped the transport coroutine scope and then waited for the scope job to finish. On the JVM, a `Source.readAtMostTo()` call backed by a blocking Java stream does not necessarily return when the coroutine is cancelled. If the read loop stayed blocked, `close()` could wait indefinitely.

## Changes

- Close the stdin source during stdio client shutdown before waiting for transport coroutines to finish.
- Also close the optional stderr source during shutdown.
- Keep source-close failures non-fatal and log them at debug level.
- Add a JVM regression test with a `RawSource` that blocks in `readAtMostTo()` until `close()` is called.
- Assert the close path completes, emits `onClose`, and does not invoke `onError`.

## Scope / Risk

This is limited to `StdioClientTransport` shutdown behavior. It does not change the public API or server transports. The regression coverage is JVM-focused because the reported hang is caused by a blocking Java stream read.

## Verification

```bash
./gradlew :integration-test:jvmTest --tests 'io.modelcontextprotocol.kotlin.sdk.client.StdioClientTransportTest'
./gradlew :kotlin-sdk-client:jvmTest
./gradlew :kotlin-sdk-client:ktlintCheck :integration-test:ktlintCheck :kotlin-sdk-client:detekt :integration-test:detekt :kotlin-sdk-client:apiCheck
```

Closes #514

### DIFF
--- a/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StdioClientTransportTest.kt
+++ b/integration-test/src/jvmTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StdioClientTransportTest.kt
@@ -7,9 +7,13 @@ import io.modelcontextprotocol.kotlin.sdk.types.McpException
 import io.modelcontextprotocol.kotlin.test.utils.createSleepyProcessBuilder
 import io.modelcontextprotocol.kotlin.test.utils.createTeeProcessBuilder
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.withTimeout
+import kotlinx.coroutines.withTimeoutOrNull
+import kotlinx.io.Buffer
+import kotlinx.io.RawSource
 import kotlinx.io.asSink
 import kotlinx.io.asSource
 import kotlinx.io.buffered
@@ -18,10 +22,12 @@ import org.junit.jupiter.api.Timeout
 import org.junit.jupiter.api.assertThrows
 import org.junit.jupiter.api.condition.DisabledOnOs
 import org.junit.jupiter.api.condition.OS
+import java.util.concurrent.CountDownLatch
 import java.util.concurrent.TimeUnit
 import kotlin.concurrent.atomics.AtomicBoolean
 import kotlin.concurrent.atomics.ExperimentalAtomicApi
 import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
 import kotlin.test.fail
 import kotlin.time.Duration.Companion.seconds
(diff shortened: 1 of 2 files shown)


devcrocod (approved):
lgtm!

PR #870, opened 2026-07-05
fix(core): apply the request timeout while awaiting the response

### LINKED ISSUE #869: `Protocol.request` timeout only bounds sending the request, not awaiting the response
**Describe the bug**

In `Protocol.request()`, `withTimeout(timeout)` wraps only the `transport.send(...)` call; awaiting the response (`result.await()`) happens outside the block. When the peer never delivers a response, the request therefore stays suspended past `RequestOptions.timeout` (default `DEFAULT_REQUEST_TIMEOUT` = 60s).

The KDoc for `RequestOptions.timeout` states "If exceeded, a `McpException` with code `RequestTimeout` is raised from `Protocol.request`", which differs from the current behavior.

Affected version: `0.14.0`; observed on `main` at 420b64d.

**To Reproduce**

1. Connect a `Protocol` to a `Transport` that accepts outgoing messages but never delivers a response.
2. Call `request()` with `RequestOptions(timeout = 100.milliseconds)`.
3. Drive the test scheduler until idle; a correct implementation should advance to the request timeout and complete the request.
4. Expected: the request fails with an `McpException` (code `RequestTimeout`). Actual: on `main`, the request remains suspended because the response wait is outside the timeout scope.

<details>
<summary>Runnable reproducer (placed in kotlin-sdk-core commonTest)</summary>

```kotlin
@OptIn(ExperimentalCoroutinesApi::class)
class RequestTimeoutReproTest {
    @Test
    fun `request fails with RequestTimeout when the response never arrives`() = runTest {
        val protocol = object : Protocol(null) {
            override fun assertCapabilityForMethod(method: Method) {}
            override fun assertNotificationCapability(method: Method) {}
            override fun assertRequestHandlerCapability(method: Method) {}
        }
        val transport = object : Transport {
            override suspend fun start() {}
            // accepts the message, never delivers a response
            override suspend fun send(message: JSONRPCMessage, options: TransportSendOptions?) {}
            override suspend fun close() {}
            override fun onClose(block: () -> Unit) {}
            override fun onError(block: (Throwable) -> Unit) {}
            override fun onMessage(block: suspend (JSONRPCMessage) -> Unit) {}
        }
        protocol.connect(transport)

        var outcome: McpException? = null
        val inFlight = launch {
            try {
                protocol.request<EmptyResult>(
                    CustomRequest(method = Method.Custom("example"), params = null),
                    RequestOptions(timeout = 100.milliseconds),
                )
            } catch (e: McpException) {
                outcome = e
            }
        }

        try {
            advanceUntilIdle()

            // On current `main` the request is still suspended here, so this fails fast.
            assertTrue(inFlight.isCompleted, "request should complete when RequestOptions.timeout elapses")
            assertEquals(RPCError.ErrorCode.REQUEST_TIMEOUT, outcome?.code)
        } finally {
            inFlight.cancelAndJoin()
        }
    }
}
```

</details>

**Expected behavior**

Once the timeout elapses, the request fails with the `McpException` (code `RequestTimeout`) documented in the KDoc.

For reference, in the TypeScript SDK the request timeout is [armed right before the send](https://github.com/modelcontextprotocol/typescript-sdk/blob/ddba550fa91cd5226d994ddf63f1147fb707131d/packages/core-internal/src/shared/protocol.ts#L1554-L1559) with a handler that rejects the request when it fires, and stays active until the request settles — so it bounds the response wait.

**Logs**

N/A (covered by the reproducer above).

**Additional context**

The relevant code (`withTimeout` wraps only the send; `result.await()` is outside):

https://github.com/modelcontextprotocol/kotlin-sdk/blob/420b64d1bb2d35e86adf544499e6748ee314ec72/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt#L503-L521

The KDoc:

https://github.com/modelcontextprotocol/kotlin-sdk/blob/420b64d1bb2d35e86adf544499e6748ee314ec72/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt#L86-L88

Note that even when the send itself times out today, the `McpException` with `RPCError.ErrorCode.REQUEST_TIMEOUT` constructed inside the catch block is only completed into a deferred that is no longer awaited; what reaches the caller is the `TimeoutCancellationException` rethrown at the end.

A minimal fix moves `result.await()` inside the `withTimeout` block and surfaces the documented `McpException` (code `RequestTimeout`) on timeout, with a regression test. Since the timeout currently never fires while awaiting the response, the exception-type change only affects the path where the send itself blocks. PR to follow.

rea9r:
Make `Protocol.request()` apply `RequestOptions.timeout` to the whole send-and-response-wait sequence, surface the documented `McpException` (code `RequestTimeout`) when it fires, and remove each request's response/progress handlers on every exit path.

closes #869

## Motivation and Context

`withTimeout` wrapped only `transport.send(...)`; `result.await()` was outside the block, so a request against a peer that never delivers a response stayed suspended past `RequestOptions.timeout`. Details in #869.

## How Has This Been Tested?

Added `ProtocolTimeoutTest` (7 cases) covering the timeout, caller-cancellation, normal-completion, and send-failure paths. They pass locally; reverting this change makes 6 of the 7 fail on `main` (only the normal-completion case still passes), confirming the tests pin the new behavior. Locally, the JVM test suites of all modules, the `integration-test` `ClientTest` suite, and `apiCheck` / `ktlintCheck` / `detekt` also pass.

## Breaking Changes

No public API changes (`apiCheck` passes). For the newly-covered response-wait path, this turns an indefinite suspension into the documented `McpException` (code `RequestTimeout`). The exception-type change from `TimeoutCancellationException` to that `McpException` (with the underlying `TimeoutCancellationException` attached as `cause`) only affects the pre-existing path where `send` itself blocked past the timeout.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

- Moving `result.await()` inside `withTimeout` means an enclosing scope's cancellation (e.g. a caller's own `withTimeout`) can also surface in the catch as a `TimeoutCancellationException`. `currentCoroutineContext().ensureActive()` at the top of the catch rethrows it, so a genuine cancellation is not converted into a request-timeout error (covered by the existing caller-side-timeout integration test, `should handle request timeout`).
- A `finally` now removes the request's response/progress handler entries on every exit path (completion, timeout, caller cancellation, send failure). This closes a pre-existing leak when the caller cancels mid-await or the send fails, and keeps the cancellation guard above from leaking on the enclosing-`withTimeout` path.
- On caller-initiated cancellation the request now propagates the cancellation and cleans up without attempting to send `notifications/cancelled` from the cancelled coroutine. A best-effort courtesy notification from `NonCancellable` could be considered separately.
- Also updates one stale test comment (`StreamableHttpClientTransportTest`) that described the previous exception type.

### DIFF
--- a/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/streamable/http/StreamableHttpClientTransportTest.kt
+++ b/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/streamable/http/StreamableHttpClientTransportTest.kt
@@ -471,7 +471,7 @@ class StreamableHttpClientTransportTest {
         )
 
         try {
-            // Real time-keeping is needed; otherwise Protocol will always throw TimeoutCancellationException in tests
+            // Real time-keeping is needed; otherwise Protocol will always fail the request with a timeout in tests
             val mcpException = assertFailsWith<McpException>(
                 message = "Expected client.connect to fail on invalid JSON response",
             ) {
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
@@ -32,6 +32,8 @@ import kotlinx.collections.immutable.persistentMapOf
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Deferred
 import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.withTimeout
 import kotlinx.serialization.json.JsonObject
 import kotlinx.serialization.json.JsonPrimitive
@@ -483,9 +485,6 @@ public abstract class Protocol(@PublishedApi internal val options: ProtocolOptio
         }
 
         val cancel: suspend (Throwable) -> Unit = { reason: Throwable ->
-            _responseHandlers.update { current -> current.remove(jsonRpcRequestId) }
-            _progressHandlers.update { current -> current.remove(jsonRpcRequestId) }
-
             val notification = CancelledNotification(
                 params = CancelledNotificationParams(
                     requestId = jsonRpcRequestId,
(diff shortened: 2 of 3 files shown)


devcrocod:
Thanks a lot, this was spot on
Meanwhile main was reworked in #884, which rewrote `Protocol.request()` and independently fixed this. That's also why this PR now conflicts
So #870 is superseded and I'll close it, along with #869. Really appreciate the contribution, hope to see more from you.

rea9r:
No worries at all. Glad #884 covers it and the fix is landing either way.
I learned a lot reading that rework. Thanks for the kind words, and I'll look for other areas to help with!

PR #872, opened 2026-07-05
close stateless streamable http session after request completion

### LINKED ISSUE #786: Stateless Streamable HTTP: server sessions are registered but never removed
# Stateless Streamable HTTP: server sessions are registered but never removed

## Summary

When serving MCP over the stateless Streamable HTTP transport
(`Application.mcpStatelessStreamableHttp { ... }`), every incoming POST request
registers a new `ServerSession` but never removes it. The session registry and its
notification-service subscriptions grow without bound across requests, since the
stateless endpoint creates a fresh `Server`/transport/session per request and never
closes them.

## Affected version

- `io.modelcontextprotocol:kotlin-sdk` **0.12.0**, and current `main`
  (verified at commit `37ee423`, 31 commits past the 0.12.0 tag — no relevant fix
  in between).

## Environment

- JVM 21, Kotlin 2.3.x, Ktor 3.5.0 (Netty)
- Transport: `Application.mcpStatelessStreamableHttp`, POST `/mcp`, `enableJsonResponse = true`

## Steps to reproduce

Start a minimal stateless server exposing one tool, then send several `tools/call`
requests:

```bash
for n in 1 2 3 4 5; do
  curl -s -H "Content-Type: application/json" \
       -H "Accept: application/json, text/event-stream" \
       -X POST http://localhost:8080/mcp \
       -d "{\"jsonrpc\":\"2.0\",\"id\":$n,\"method\":\"tools/call\",\"params\":{\"name\":\"ping\",\"arguments\":{}}}"
done
```

## Actual behavior

`ServerSessionRegistry` logs one `Adding session` per request and never logs
`Removing session`; the registered-session count rises monotonically. The per-request
`block` is also re-invoked each time (a new `Server` and its tools are rebuilt per
request).

| Log line          | Count over 5 requests |
|-------------------|-----------------------|
| `Adding session`  | 5                     |
| `Removing session`| 0                     |

## Expected behavior

A session created to serve a stateless request should not outlive that request:
`Removing session` balances `Adding session`, subscriptions are released, and the
registry does not grow across requests.

## Where it happens

- `mcpStatelessStreamableHttpEndpoint` in `kotlin-sdk-server/.../server/KtorServer.kt`
  creates the transport, calls `server.createSession(transport)`, runs
  `transport.handleRequest(...)`, and returns.
- Session removal and unsubscription are wired to `session.onClose` in
  `Server.createSession` (`Server.kt`), but nothing closes the transport/session on
  the stateless request path, so `onClose` never fires.

## Related issues

- #562 — Ktor extensions for `StreamableHttpServerTransport` (where the stateless endpoint lives)
- #297 — established that sessions must be removed from the registry on disconnect to avoid OOM
- #296 — re-creating the `Server` (and its tools) per transport/request

KlyneChrysler:
fixes #786

stateless streamable http endpoint never closed its per request session, so the session registry and notification subscriptions grew forever. now the session is closed in a finally block after each request, which triggers the existing cleanup. also removed the per request server.onClose logging callback that accumulated on shared servers.

added tests proving sessions are removed after normal and rejected requests. jvmTest, ktlint, detekt, apiCheck all pass. no public api changes.

### DIFF
--- a/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
+++ b/kotlin-sdk-server/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/server/KtorServer.kt
@@ -380,11 +380,17 @@ private suspend fun RoutingContext.mcpStatelessStreamableHttpEndpoint(
     logger.info { "New stateless StreamableHttp connection established without sessionId" }
 
     val server = block()
-    server.onClose { logger.info { "Server connection closed without sessionId" } }
-    server.createSession(transport)
+    val session = server.createSession(transport)
 
-    transport.handleRequest(null, this.call)
-    logger.debug { "Server connected to transport without sessionId" }
+    try {
+        transport.handleRequest(null, this.call)
+        logger.debug { "Server connected to transport without sessionId" }
+    } finally {
+        // A stateless session serves exactly one request: close it so the server's
+        // session registry and notification subscriptions do not grow unboundedly.
+        session.close()
+        logger.debug { "Stateless session closed after request completion" }
+    }
 }
 
 private suspend fun RoutingContext.mcpPostEndpoint(transportManager: TransportManager<SseServerTransport>) {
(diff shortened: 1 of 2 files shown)


devcrocod (approved):
thanks for contributing!
lgtm

PR #913, opened 2026-07-23
fix(core): disambiguate task result responses

### LINKED ISSUE #601: type-unsafe cast in Protocol result deserialization

jstar0:
## Summary

Fix task response deserialization so `tasks/result` preserves the original result object even when its shape overlaps another known result type, while `tasks/get` and `tasks/cancel` continue to return typed task state.

## Root Cause

JSON-RPC responses contain an ID and result but not the original request method. The response serializer therefore classifies `RequestResult` from JSON shape alone. That is insufficient for task APIs: `tasks/result` returns the original request's result object, so a valid payload can look like `GetTaskResult`, `CallToolResult`, or an extension-defined result.

The payload could consequently be decoded into the wrong runtime subtype or rejected before `Protocol.request()` correlated it with the request method.

## Changes

- Preserve the raw JSON-RPC response result during message decoding.
- Resolve `tasks/get`, `tasks/cancel`, and `tasks/result` using the correlated request method before completing the request.
- Re-encode directly constructed transport responses when raw wire JSON is unavailable.
- Keep the existing public request API and generated `JSONRPCResponse` serializer unchanged.
- Add regressions for task-state-shaped, tool-shaped, unknown, custom-method, and direct-transport task results, plus JSON-RPC message serialization compatibility.

## Scope and Risk

The change is limited to task response result selection. Existing non-task methods keep their current content-based result type. Unknown object-shaped responses can now reach request correlation as raw task payloads; the original result is returned unchanged only for `tasks/result`.

`GetTaskPayloadResult` is currently object-backed, so array and scalar task payloads remain outside this change.

The existing request API erases `T`, so its final internal completion cast remains. This change makes the task result's runtime subtype correct before that boundary. Removing the cast itself would require changing the public request contract, which was the breaking approach rejected in #717.

## Verification

```bash
./gradlew --no-parallel :kotlin-sdk-core:jvmTest :kotlin-sdk-core:apiCheck :kotlin-sdk-core:ktlintCheck :kotlin-sdk-core:detekt
./gradlew --no-parallel :kotlin-sdk-core:jsTest :kotlin-sdk-core:wasmJsTest
./gradlew --no-parallel :kotlin-sdk-core:macosArm64Test :kotlin-sdk-core:iosSimulatorArm64Test
```

Closes #601

### DIFF
--- a/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
+++ b/kotlin-sdk-core/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/shared/Protocol.kt
@@ -4,6 +4,8 @@ import io.github.oshai.kotlinlogging.KotlinLogging
 import io.modelcontextprotocol.kotlin.sdk.types.CancelledNotification
 import io.modelcontextprotocol.kotlin.sdk.types.CancelledNotificationParams
 import io.modelcontextprotocol.kotlin.sdk.types.EmptyResult
+import io.modelcontextprotocol.kotlin.sdk.types.GetTaskPayloadResult
+import io.modelcontextprotocol.kotlin.sdk.types.GetTaskResult
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCEmptyMessage
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCError
 import io.modelcontextprotocol.kotlin.sdk.types.JSONRPCNotification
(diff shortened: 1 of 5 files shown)


devcrocod:
Thanks for looking into this. Unfortunately, I don't think this approach will work.

The main issue is that `GetTaskPayloadResult` can be any JSON object, so we can't reliably distinguish it from `GetTaskResult` by shape. It also breaks for Streamable HTTP, since copying the response there drops the raw result.
I think the better fix is to keep the raw JSON and deserialize it based on the method when correlating the response, similar to how request deserialization already works.
I’m going to close this for now, but happy to review a PR along those lines. Thanks again!

PR #953, opened 2026-08-11
feat(server): allow decorating Ktor transports

### LINKED ISSUE #921: Ktor SSE / StreamableHttp extensions build their transports internally, with no hook to wrap them
**Affected version / component:** `io.modelcontextprotocol:kotlin-sdk` 0.14.0 — `server/KtorServer.kt`

The `Application.mcp { }` (SSE) and `Application.mcpStreamableHttp { }` extensions construct the transport and create the session entirely inside the extension. The user block only returns a `Server`, so there is no point at which caller code can wrap, decorate, or intercept the `Transport` before it is used. That rules out any cross-cutting transport behavior — logging, metrics, back-pressure, custom framing — on HTTP sessions, even though it is straightforward for transports you build yourself.

The SSE endpoint is representative:

```kotlin
private suspend fun ServerSSESession.mcpSseEndpoint(..., block: ServerSSESession.() -> Server) {
    val transport = mcpSseTransport(postEndpoint, transportManager, maxRequestBodySize) // SseServerTransport(...)
    val server = block()
    server.onClose { ... }
    server.createSession(transport)
    awaitCancellation()
}
```

`mcpStreamableHttp` has the same shape, with `StreamableHttpServerTransport(configuration)` followed by `server.createSession(transport)`. Because the block returns a `Server`, the caller never sees the `Transport` and cannot substitute a delegating wrapper. `mcpStreamableHttp { }` does accept a `serverFactory` parameter, but it only supplies the `Server`; it does not expose or allow wrapping the transport, so it does not close this gap.

(Surfacing the JSON-RPC request id to handlers, which might otherwise motivate a transport hook, is being addressed separately upstream, so it is out of scope here.)

### Suggested fix

Expose a transport factory/interceptor hook — let the `mcp` / `mcpStreamableHttp` setup wrap each `Transport` before `createSession`, or route these sessions through a user-controllable `connect()` path so a subclassed server can intercept them.

### Workaround

The HTTP/SSE transport classes (`SseServerTransport`, `StreamableHttpServerTransport`) and their `handleRequest`/`handlePostMessage` methods are public, so you can stand up your own Ktor routes, construct the transport, wrap it, and call `Server.createSession(wrappedTransport)` directly. The cost is real: the extension wiring you have to replicate — `TransportManager`, the session-id lifecycle (`setOnSessionInitialized`/`setOnSessionClosed`), DNS-rebinding protection, and content negotiation — is private, roughly a hundred lines that then have to track SDK internals across upgrades.

### Impact

A transport hook would bring logging, metrics, back-pressure, and custom framing to HTTP/SSE without reimplementing the extension. Today, parity between stdio and HTTP transports means duplicating that internal wiring.

ECD5A:
## Motivation and Context

Closes #921.

The Ktor SSE and Streamable HTTP extensions create their transports internally, so applications cannot add transport-level logging, metrics, back-pressure, or framing without reimplementing the SDK's private route and session wiring.

This adds opt-in `transportDecorator` overloads for the SSE `mcp` extensions and `mcpStreamableHttp`. Each decorator receives the concrete SDK transport and returns the `Transport` passed to `Server.createSession`. Existing overloads delegate through an identity decorator, preserving current source and binary APIs.

## How Has This Been Tested?

- Added Ktor integration tests for `Application.mcp`, `Route.mcp(path)`, and `Application.mcpStreamableHttp`.
- The tests wrap each concrete transport, verify the wrapper is started, and exercise the normal SSE ping or Streamable HTTP initialize flow.
- Ran the complete `kotlin-sdk-server` JVM test suite.
- Ran `ktlintCheck`, `detekt`, and `apiCheck` for the repository.

## Breaking Changes

None. Existing overloads and their generated JVM signatures remain unchanged; the new overloads are additive.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

The concrete transport remains responsible for Ktor request handling and session bookkeeping. A decorator must delegate lifecycle methods and callback registration to that supplied transport; this requirement is documented on each new overload.


devcrocod:
Thanks for this!
I am closing this PR because this patch only partially solves the problem. `createSession` is called in different places, which means it would need to add even more overloads and increase the number of public methods. This limits future SDK stabilization. The problem still exists, but since there are no real use cases yet, this issue is not blocking anything.
Therefore, a solution requiring long-term API changes does not seem appropriate at the moment

PR #957, opened 2026-08-16
fix: reject batch requests in streamable HTTP transport

ez-lbz:
MCP has removed support for JSON-RPC batches: a POST body must be a single JSON-RPC message. The streamable HTTP server currently decodes a `JsonArray` body into a `List<JSONRPCMessage>` and replies with an array response.

This change rejects batch request bodies with `400 Bad Request` (JSON-RPC error code `-32600`), mirroring the handling of other invalid payloads, and replaces the batch-specific tests with one asserting the rejection.


devcrocod:
Thanks for the contribution!
The specification doesn't say that the server should reject batch requests. Also, I checked the ts sdk and it still supports them, so it is better to keep this to ensure compatibility between different clients and servers

PR #958, opened 2026-08-16
fix: reject cross-origin endpoint events in SSE client

ez-lbz:
In the HTTP+SSE transport, the server's `endpoint` event tells the client where to POST subsequent JSON-RPC messages. The client currently accepts a full `http(s)://` URL as-is, so a compromised server could redirect all subsequent traffic — including auth headers — to an attacker-controlled host. The TypeScript and Python clients already reject endpoints whose origin differs from the connection origin.

This change rejects full-URL endpoints whose origin (scheme, host, and port, with default ports normalized) does not match the SSE connection's origin: the endpoint future completes exceptionally and the transport fails to start. Relative and root-relative paths keep their existing resolution, and same-origin full URLs keep their existing behavior.

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
@@ -11,6 +11,7 @@ import io.ktor.client.request.setBody
 import io.ktor.client.statement.bodyAsText
 import io.ktor.http.ContentType
 import io.ktor.http.HttpHeaders
+import io.ktor.http.Url
 import io.ktor.http.append
 import io.ktor.http.isSuccess
 import io.ktor.http.protocolWithAuthority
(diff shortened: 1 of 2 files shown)


kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
```diff
@@ -163,6 +173,13 @@ public class SseClientTransport(
         }
     }
 
+    /**
+     * Returns true when [this]'s origin (scheme, host, and port) matches the SSE connection's origin.
+     * The comparison uses the authority string, so default ports are ignored (e.g. `http://host:80`
+     * and `http://host` are considered the same origin).
+     */
+    private fun Url.hasSameOrigin(): Boolean = protocolWithAuthority == origin
```
devcrocod, line 181: A raw string is compared, which can also include UserAndPassword. This can cause rejection in two cases, when credentials are present or when there is a case mismatch.

Also, why is this moved to a separate function? It is not used anywhere else, and it also increases complexity

kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/SseClientTransport.kt
```diff
@@ -142,12 +143,21 @@ public class SseClientTransport(
...
+     * Uses full URLs as-is, but rejects those whose origin differs from the SSE connection origin,
+     * treats absolute paths as origin-relative, and relative paths as relative to [baseUrl].
      */
     private fun handleEndpoint(eventData: String) {
         try {
             val endpointUrl = if (eventData.startsWith("http://") || eventData.startsWith("https://")) {
+                val endpointOrigin = Url(eventData)
+                if (!endpointOrigin.hasSameOrigin()) {
+                    val error = IllegalArgumentException(
+                        "Endpoint origin ${endpointOrigin.protocolWithAuthority} does not match connection origin $origin",
```
devcrocod, line 155: credentials might leak into error messages

kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/sse/SseClientTransportTest.kt
devcrocod, line 1: I can't find a test for port mismatch, which seems like a frequent use case

devcrocod (changes requested):
there are a few critical comments, please take a look
also note that ktlint failed on ci

PR #965, opened 2026-08-20
feat: support suspending close callbacks

### LINKED ISSUE #919: onClose callbacks are synchronous, so suspending cleanup has nowhere to run
**Affected version / component:** `io.modelcontextprotocol:kotlin-sdk` 0.14.0 — `shared/Transport.kt`, `shared/AbstractTransport.kt`, `server/ServerSession.kt`, `server/Server.kt`

Every close callback in the server stack is a plain `() -> Unit`. `Transport.onClose`, `ServerSession.onClose`, and `Server.onClose` all take a non-suspending block, and `Server.close()` invokes the registered chain synchronously. That leaves nowhere to run suspending teardown — closing pooled resources, or cancelling and then *joining* a child `CoroutineScope` — without wrapping it in `runBlocking`, which can stall or deadlock the close path.

The contract is synchronous at each layer. `AbstractTransport` composes the registered blocks as `() -> Unit`, and `invokeOnCloseCallback()` runs them inline inside `runCatching`, swallowing exceptions. `ServerSession.onClose` chains the same way, and its `onClose()` hook fires the chain synchronously. `Server.onClose` does likewise, and `Server.close()` calls it directly:

```kotlin
public suspend fun close() {
    notificationService.close()
    sessions.forEach { (_, session) -> session.close() }
    _onClose() // synchronous; no suspension point for teardown
}
```

`close()` is itself a suspend function, so the synchronous callback is a limitation of the API contract rather than something the implementation forces.

For context, #778 (shipped in 0.13.0) made the stdio transport's own `close()` suspending, with a graceful drain of in-flight outbound messages, and moved message handlers to `Dispatchers.Default`. It did not touch the user-facing close callbacks, which are still non-suspending everywhere, and there is no `awaitClose()` in 0.14.0.

### Suggested fix

Offer a suspending close path — for example an `onClose(block: suspend () -> Unit)` overload, or a `suspend` close hook that `Server.close()` and session/transport teardown await. Either lets a server with suspending resources shut down deterministically instead of bridging through `runBlocking`.

### Workaround

Keep the `onClose` block limited to synchronous work (flipping flags, cancelling a scope without joining it) and move all suspending teardown into a separate `suspend` function that callers invoke and await after close. Alternatively, treat the callback purely as a signal that resumes a coroutine, and run the suspending cleanup outside it:

```kotlin
suspend fun awaitClose(session: ServerSession) = suspendCoroutine { cont ->
    session.onClose { cont.resume(Unit) }
}
// awaitClose(session); mySuspendingCleanup()
```

Both work, but they push lifecycle management onto every caller, which is easy to get wrong.

original4422:
## Summary

- change transport, protocol, session, and server close callbacks to suspending callbacks that are awaited by their close paths
- preserve callback registration order, `AbstractTransport`'s exactly-once behavior, and its existing exception suppression while allowing cleanup to finish after transport-scope cancellation
- bridge WebSocket and server-SSE completion notifications through lifecycle-owned watcher jobs, including an explicit SSE close path that does not wait for the session Job to finish
- update public KDoc and API dumps for the intentional breaking API change

## Testing

- `./gradlew :kotlin-sdk-core:jvmTest :kotlin-sdk-client:jvmTest :kotlin-sdk-server:jvmTest :kotlin-sdk-testing:jvmTest`
- `./gradlew ktlintCheck apiCheck detekt`
- `./gradlew apiDump`
- compiled an out-of-repository consumer that calls `delay()` inside `Transport.onClose`

Fixes #919

### DIFF
--- a/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StdioClientTransport.kt
+++ b/kotlin-sdk-client/src/commonMain/kotlin/io/modelcontextprotocol/kotlin/sdk/client/StdioClientTransport.kt
@@ -258,7 +258,7 @@ public class StdioClientTransport @JvmOverloads public constructor(
         }
     }
 
-    private fun sendOutboundMessage(message: JSONRPCMessage, sink: Sink, mainScope: CoroutineScope) {
+    private suspend fun sendOutboundMessage(message: JSONRPCMessage, sink: Sink, mainScope: CoroutineScope) {
         try {
             val json = serializeMessage(message)
             sink.writeString(json)
@@ -285,7 +285,7 @@ public class StdioClientTransport @JvmOverloads public constructor(
         }
     }
 
-    private fun CoroutineScope.stopProcessing(reason: String, cause: Throwable? = null) {
+    private suspend fun CoroutineScope.stopProcessing(reason: String, cause: Throwable? = null) {
         sendChannel.close() // Stop accepting new messages
         invokeOnCloseCallback()
         cancel(reason, cause) // cancel current coroutine context
--- a/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ClientAssertCapabilityTest.kt
+++ b/kotlin-sdk-client/src/commonTest/kotlin/io/modelcontextprotocol/kotlin/sdk/client/ClientAssertCapabilityTest.kt
@@ -135,7 +135,7 @@ class ClientAssertCapabilityTest {
      */
     private class CapabilitiesTransport(private val serverCapabilities: ServerCapabilities) : Transport {
         private var onMessageBlock: (suspend (JSONRPCMessage) -> Unit)? = null
-        private var onCloseBlock: (() -> Unit)? = null
+        private var onCloseBlock: (suspend () -> Unit)? = null
 
         override suspend fun start() = Unit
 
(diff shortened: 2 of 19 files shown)


original4422:
Hi @devcrocod, when you have a chance, would you mind reviewing this PR and letting me know whether any changes are needed? Thanks!

original4422:
Hi @devcrocod, following up since the last review request—when you have a chance, could you please take another look and let me know whether this is ready to merge or needs changes? Thanks!

original4422:
Hi @devcrocod, the repository is active again and this PR has been waiting since the last review request. When convenient, could you please check whether it is ready to merge or needs any changes? Thanks!

original4422:
Hi @devcrocod, when convenient, could you please check whether this PR is ready to merge or needs any changes? Thank you!

original4422:
Hi @devcrocod, when convenient, could you please check whether this PR can be merged or needs any changes? Thank you!

devcrocod:
Thanks for the work here
I'm closing it. This isn't the best direction for API and design at all. And changes have problems on top of that. 

- `onClose` is a notificaiton, not a finalizer. An application's pools and subscriptions aren't the transport's business, and awaiting them inside its close path makes our shutdown depend on user code.
- Adding `suspend` forces changes to other dependent functions and also complicates the API. One signature colors the whole teardown graph
- `StreamableHttpServerTransport` was missed entirely. It still invokes the callback under `streamMutex`, which a suspending callback now pins non-cancellable. None of it is covered
- No deprecation step

